PR #5, opened 2021-05-14
Ignore .idea directory

fredx30:
This ignores the .idea directory that gets created when opening the project with the Jetbrains GoLand IDE.

I have a thoery that the .gitignore file was created on windows without the `git config --global core.autocrlf true` set.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,6 +6,7 @@ Dockerfile
 
 .wharf-ci.yml
 docs
+.idea
 
 package.json
 package-lock.json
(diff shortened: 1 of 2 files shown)


Pikabanga (approved):

jilleJr (approved):
> I have a thoery that the .gitignore file was created on windows without the git config --global core.autocrlf true set.

Yep, some files in these repos has CRLF line endings pushed. I'm partially of blame and most files have been fixed, but I guess the dotfiles has not.

PR #26, opened 2021-06-01
Add endpoint PUT /token

fredx30:
This adds a code path for PUT for `/api/token`

This goes together with a few other PRs:
- The implementable interface https://github.com/iver-wharf/wharf-api-client-go/pull/8
- (This) The PUT /api/token https://github.com/iver-wharf/wharf-api/pull/26
- The PUT /api/provider https://github.com/iver-wharf/wharf-api/pull/28
- The gtihub-provider changes that require the rest https://github.com/iver-wharf/wharf-provider-github/pull/10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 - Added endpoint `PUT /provider` as an idempotent way of creating a
   provider. (#28)
 
+- Add endpoint `PUT /token` as an idempotent way of creating a
+  token. (#26)
+
 - Added environment var for setting bind address and port. (#29)
 
 - Fixed missing `failed` field from `main.TestsResults` in
--- a/token.go
+++ b/token.go
@@ -74,6 +74,7 @@ func (m TokenModule) getTokenHandler(c *gin.Context) {
 	} else if err != nil {
 		problem.WriteDBReadError(c, err, fmt.Sprintf(
 			"Failed fetching token by ID %d from database.", tokenID))
+		return
 	}
 
 	c.JSON(http.StatusOK, token)
--- a/token.go
+++ b/token.go
@@ -124,7 +124,6 @@ type TokenWithProviderID struct {
 // postTokenHandler godoc
 // @summary Add token to database.
 // @description Add token to database. Provider in post object has to exists or should be empty.
-// @description Provider will has to be updated token ID during this operation.
 // @tags token
 // @accept json
 // @produce json
--- a/token.go
+++ b/token.go
@@ -200,6 +200,11 @@ func (m TokenModule) putTokenHandler(c *gin.Context) {
 		return
 	}
 	var placedToken Token
-	m.Database.Where(token).FirstOrCreate(&placedToken)
+	if err := m.Database.Where(token).FirstOrCreate(&placedToken).Error; err != nil {
+		problem.WriteInvalidBindError(c, err, fmt.Sprintf(
+			"Failed fetch or create on token with token:%q and username:%q.",
+			token.Token, token.UserName))
+		return
+	}
 	c.JSON(http.StatusAccepted, placedToken)
 }
--- a/token.go
+++ b/token.go
@@ -27,6 +27,7 @@ func (m TokenModule) Register(g *gin.RouterGroup) {
 	{
 		token.GET("/:tokenid", m.getTokenHandler)
 		token.POST("", m.postTokenHandler)
+		token.PUT("", m.putTokenHandler)
 	}
 }
 
@@ -177,4 +178,30 @@ func (m TokenModule) postTokenHandler(c *gin.Context) {
 	}
 
 	c.JSON(http.StatusCreated, token)
+}
+
+// postTokenHandler godoc
+// @summary Put token in database.
+// @description Put token in database. Provider in post object has to exists or should be empty.
+// @description Provider will has to be updated token ID during this operation.
+// @tags token
+// @accept json
+// @produce json
+// @param token body TokenWithProviderID _ "token object"
+// @success 201 {object} Token
+// @failure 400 {object} problem.Response "Bad request"
+// @failure 401 {object} problem.Response "Unauthorized or missing jwt token"
+// @failure 404 {object} problem.Response "Referenced provider not found"
+// @failure 502 {object} problem.Response "Database is unreachable"
+// @router /token [put]
+func (m TokenModule) putTokenHandler(c *gin.Context) {
+	var token Token
+	err := c.ShouldBindJSON(&token)
+	if err != nil {
+		problem.WriteInvalidBindError(c, err, "One or more parameters failed to parse when reading the request body.")
+		return
+	}
+	var placedToken Token
+	m.Database.Where(token).FirstOrCreate(&placedToken)
+	c.JSON(http.StatusAccepted, placedToken)
 }
--- a/token.go
+++ b/token.go
@@ -196,8 +196,7 @@ func (m TokenModule) postTokenHandler(c *gin.Context) {
 // @router /token [put]
 func (m TokenModule) putTokenHandler(c *gin.Context) {
 	var token Token
-	err := c.ShouldBindJSON(&token)
-	if err != nil {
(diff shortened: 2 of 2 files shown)


jilleJr (changes requested):

token.go
```diff
@@ -178,3 +179,29 @@ func (m TokenModule) postTokenHandler(c *gin.Context) {
...
+// @router /token [put]
+func (m TokenModule) putTokenHandler(c *gin.Context) {
+	var token Token
+	err := c.ShouldBindJSON(&token)
+	if err != nil {
+		problem.WriteInvalidBindError(c, err, "One or more parameters failed to parse when reading the request body.")
+		return
+	}
+	var placedToken Token
+	m.Database.Where(token).FirstOrCreate(&placedToken)
```
jilleJr, line 205: Have not tackled this issue that CodeQL is warning about in this repo yet. But it seems it wants us to do something like this instead:

```suggestion
	m.Database.
		Where(&token, tokenFieldToken, tokenFieldUserName).
		FirstOrCreate(&placedToken)
```

Based on recommendation from https://gorm.io/docs/security.html

Not sure though, as the security warning only shows recommendations on regular `database/sql` usage, and the GORM security page only shows info on using SQL column names directly like `.Where("token = ?", token.Token)`. Want to refrain from using column names in strings, and `.Where(tokenFieldToken + "= ?", token.Token)` is just overly verbose

https://github.com/iver-wharf/wharf-api/security/code-scanning/14?query=ref%3Arefs%2Fpull%2F26%2Fmerge

jilleJr (changes requested):

token.go
fredx30: As far as my tests and understanding on this goes-
The sql injection mentioned in the gorm security doc you linked is only available when gorm isnt given the opportinuty to split the collumn name from the value its being set to. 

This is not relevant in our case because the first object (int this case token) is adequately constructed. The named values are split by gorm. In this case in a key:value pairing.

The resulting sql query from both with/without the sugestion is as follows:
```bash
[10.150ms] [rows:1] INSERT INTO "token" ("token","user_name") VALUES ('sample-token\'; drop table branch;','') RETURNING "token_id"
[GIN] 2021/06/02 - 17:12:08 | 202 |         3m54s |      172.19.0.1 | PUT      "/api/token"
[GIN] 2021/06/02 - 17:12:08 | 400 |            0s |      172.19.0.1 | POST
```
These results are created with a post to /import/github with a body containing 
```json
{
    "group": "iver-wharf",
    "project": "wharf-provider-github",
    "token": "sample-token'; drop table branch;",
    "url": "https://api.github.com"
}
```
As you can see the escape is correct in both cases.

This is also works (and escapes) correctly if the code is changed to something like 
```go
m.Database.Where("Token=?", token.Token)
```


The issue of sql injection first seems to arise when this is changed to
```go
m.Database.Where("token=" + token.Token)
```
Even in this scenario it still uses PDOs so it appears multiple sql statements are impossible but it seems we could go fishing for other friendly tokens- (tested and works)
```json
{
    "group": "iver-wharf",
    "project": "wharf-provider-github",
    "token": "'sample-token' OR token not like 'sample-token'",
    "url": "https://token.fishing.me"
}
```

Final words - If i didnt miss anything then i propse we avoid code bloat by not passing unessacery function params. If i did miss something about how this works i would love to hear more.

token.go
```diff
@@ -178,3 +179,32 @@ func (m TokenModule) postTokenHandler(c *gin.Context) {
...
+	var token Token
+	if err := c.ShouldBindJSON(&token); err != nil {
+		problem.WriteInvalidBindError(c, err, "One or more parameters failed to parse when reading the request body.")
+		return
+	}
+	var placedToken Token
+	if err := m.Database.Where(token).FirstOrCreate(&placedToken).Error; err != nil {
+		problem.WriteInvalidBindError(c, err, fmt.Sprintf(
+			"Failed fetch or create on token with token:%q and username:%q.",
+			token.Token, token.UserName))
```
jilleJr, line 206: Wrong problem template type, should be a DB write problem IMO.

And we shall never log nor report secrets.

```suggestion
		problem.WriteDBWriteError(c, err, fmt.Sprintf(
			"Failed fetch or create on token by username %q and token value.",
			token.UserName))
```

jilleJr (changes requested):

token.go
jilleJr: Nice research! I have never tested this so it's really nice to know for sure.

I've marked the CodeQL as "False positive" in https://github.com/iver-wharf/wharf-api/security/code-scanning/15?query=ref%3Arefs%2Fpull%2F26%2Fmerge

token.go
fredx30: Was blissfully unaware op the different template types. Perhaps we could report a token hash (say sha256)? It seems the usage of usernames here is unlikely so some way of identifying which token might be usefull.

token.go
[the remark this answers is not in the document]
fredx30: This continued on a comment below.

token.go
jilleJr: Hmm maybe, but then we get into the situation where should salt the hashes as well, but we don't have anything viable to salt it with. I'd say better off to just not include the token secret at all. This message is meant to be returned by the endpoint, and in the future we might want to log these problems as well.

We want to log and return as much information as possible just so that it's easier to understand the audit logs and error messages, but with the exception of secrets.

fredx30:
```go
			"Failed fetch or create on token by username %q and token(sha256sum) %q.",
			sha256.Sum256([]byte(token.Token)),token.UserName))
```

jilleJr: You're not salting the hash there. I consider this a security leak.

We don't gain much in even showing the token SHA hash other than some obscure correlation situations. I deem us better off if we don't include the token at all.

fredx30: Yeah with the salt i imagine it would be difficult to use it even just for correlations. Removed the hash code and mirrored your sugestion.

CHANGELOG.md
```diff
@@ -12,7 +12,9 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 	https://changelog.md/
 -->
 
-## v4.0.1 (WIP)
+## v4.1.0 (WIP)
+
+- Add endpoint PUT `PUT /token`. (#26)
```
jilleJr, line 17: nit: Same comment here: double wording of "PUT". Also, not a requirement, but would be nice to say a little what the new endpoint does.

```suggestion
- Added endpoint `PUT /token` as an idempotent way of creating a token. (#26)
```

token.go
```diff
@@ -178,3 +179,32 @@ func (m TokenModule) postTokenHandler(c *gin.Context) {
 
 	c.JSON(http.StatusCreated, token)
 }
+
+// postTokenHandler godoc
+// @summary Put token in database.
+// @description Put token in database. Provider in post object has to exists or should be empty.
+// @description Provider will has to be updated token ID during this operation.
```
jilleJr, line 186: These "provider" notices here are no longer relevant.

```suggestion
// @description Creates a new token if a match is not found.
```

token.go
```diff
@@ -178,3 +179,32 @@ func (m TokenModule) postTokenHandler(c *gin.Context) {
...
+// @failure 401 {object} problem.Response "Unauthorized or missing jwt token"
+// @failure 502 {object} problem.Response "Database is unreachable"
+// @router /token [put]
+func (m TokenModule) putTokenHandler(c *gin.Context) {
+	var token Token
+	if err := c.ShouldBindJSON(&token); err != nil {
+		problem.WriteInvalidBindError(c, err, "One or more parameters failed to parse when reading the request body.")
+		return
+	}
+	var placedToken Token
```
jilleJr, line 202: Same comment here about the naming. Suggest naming them equivalent to inside the `putProviderHandler`

- `token` &rarr; `inputToken`
- `placedToken` &rarr; `token`

jilleJr (changes requested):

token.go
fredx30: Taken as refactor commit where `token → inputToken` & `placedToken → token`

jilleJr (approved):

iverestefans (approved):

PR #28, opened 2021-06-03
Add endpoint PUT /provider

fredx30:
This adds a code path for PUT for /api/provider

This goes together with a few other PRs:
- The implementable interface https://github.com/iver-wharf/wharf-api-client-go/pull/8
- The PUT /api/token https://github.com/iver-wharf/wharf-api/pull/26
- (This) The PUT /api/provider https://github.com/iver-wharf/wharf-api/pull/28
- The gtihub-provider changes that require the rest https://github.com/iver-wharf/wharf-provider-github/pull/10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 ## v4.1.0 (WIP)
 
+- Add endpoint PUT `PUT /provider`. (#28)
+
 - Added environment var for setting bind address and port. (#29)
 
 - Fixed missing `failed` field from `main.TestsResults` in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 ## v4.1.0 (WIP)
 
-- Add endpoint PUT `PUT /provider`. (#28)
+- Added endpoint `PUT /provider` as an idempotent way of creating a
+  provider. (#28)
 
 - Added environment var for setting bind address and port. (#29)
 
--- a/provider.go
+++ b/provider.go
@@ -199,17 +199,17 @@ func (m ProviderModule) postProviderHandler(c *gin.Context) {
 // @failure 502 {object} problem.Response "Database is unreachable"
 // @router /provider [put]
 func (m TokenModule) putProviderHandler(c *gin.Context) {
-	var provider Provider
-	if err := c.ShouldBindJSON(&provider); err != nil {
+	var inputProvider Provider
+	if err := c.ShouldBindJSON(&inputProvider); err != nil {
 		problem.WriteInvalidBindError(c, err, "One or more parameters failed to parse when reading the request body.")
 		return
 	}
-	var placedProvider Provider
-	if err := m.Database.Where(token).FirstOrCreate(&placedProvider).Error; err != nil {
+	var provider Provider
+	if err := m.Database.Where(token).FirstOrCreate(&provider).Error; err != nil {
 		problem.WriteDBWriteError(c, err, fmt.Sprintf(
-			"Failed fetch or create on provider with name %q.",
-			provider.Name))
+			"Failed fetch or create on inputProvider with name %q.",
+			inputProvider.Name))
 		return
 	}
-	c.JSON(http.StatusOK, placedProvider)
+	c.JSON(http.StatusOK, provider)
 }
--- a/provider.go
+++ b/provider.go
@@ -185,3 +185,31 @@ func (m ProviderModule) postProviderHandler(c *gin.Context) {
 
 	c.JSON(http.StatusCreated, provider)
 }
+
+// putProviderHandler godoc
+// @summary Put provider in database.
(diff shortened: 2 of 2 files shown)


provider.go
```diff
@@ -185,3 +185,31 @@ func (m ProviderModule) postProviderHandler(c *gin.Context) {
...
+// @failure 401 {object} problem.Response "Unauthorized or missing jwt token"
+// @failure 502 {object} problem.Response "Database is unreachable"
+// @router /provider [put]
+func (m TokenModule) putProviderHandler(c *gin.Context) {
+	var provider Provider
+	if err := c.ShouldBindJSON(&provider); err != nil {
+		problem.WriteInvalidBindError(c, err, "One or more parameters failed to parse when reading the request body.")
+		return
+	}
+	var placedProvider Provider
```
jilleJr, line 207: What's this verb "placed"? Uncommon term for variable nomenclature

Suggest renaming the variables to something like

- `provider` -> `inputProvider`
- `placedProvider` -> `dbProvider`

provider.go
```diff
@@ -185,3 +185,31 @@ func (m ProviderModule) postProviderHandler(c *gin.Context) {
 
 	c.JSON(http.StatusCreated, provider)
 }
+
+// putProviderHandler godoc
+// @summary Put provider in database.
+// @description Put provider in database.
```
jilleJr, line 191: For those uninitiated by REST verbs, maybe it'd be better to change this to something like:

```suggestion
// @description Creates a new provider if a match is not found.
```

provider.go
```diff
@@ -185,3 +185,31 @@ func (m ProviderModule) postProviderHandler(c *gin.Context) {
...
+// @failure 502 {object} problem.Response "Database is unreachable"
+// @router /provider [put]
+func (m TokenModule) putProviderHandler(c *gin.Context) {
+	var provider Provider
+	if err := c.ShouldBindJSON(&provider); err != nil {
+		problem.WriteInvalidBindError(c, err, "One or more parameters failed to parse when reading the request body.")
+		return
+	}
+	var placedProvider Provider
+	if err := m.Database.Where(token).FirstOrCreate(&placedProvider).Error; err != nil {
```
jilleJr, line 208: `.Where(token)` seems out of place here

Maybe it should be like this?

```suggestion
	if err := m.Database.Where(provider).FirstOrCreate(&placedProvider).Error; err != nil {
```

jilleJr (changes requested):

provider.go
fredx30: Fair point i think the idea was to use the past tense of PUT. Implementing `provider -> inputProvider` now.

provider.go
fredx30: Yeah that would have been awkward. Thanks!

provider.go
```diff
@@ -198,14 +198,14 @@ func (m ProviderModule) postProviderHandler(c *gin.Context) {
 // @failure 401 {object} problem.Response "Unauthorized or missing jwt token"
 // @failure 502 {object} problem.Response "Database is unreachable"
 // @router /provider [put]
-func (m TokenModule) putProviderHandler(c *gin.Context) {
+func (m ProviderModule) putProviderHandler(c *gin.Context) {
```
fredx30, line 201: Also found and fixed this one.

CHANGELOG.md
```diff
@@ -12,7 +12,9 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 	https://changelog.md/
 -->
 
-## v4.0.1 (WIP)
+## v4.1.0 (WIP)
+
+- Add endpoint PUT `PUT /provider`. (#28)
```
jilleJr, line 17: nit: Double wording of "PUT". Also, not a requirement, but would be nice to say a little what the new endpoint does.

```suggestion
- Added endpoint `PUT /provider` as an idempotent way of creating a
  provider. (#28)
```

jilleJr (approved):
YetAnotherNitpickInTheChangelog, but the endpoint looks flawless!

provider.go
```diff
@@ -185,3 +185,31 @@ func (m ProviderModule) postProviderHandler(c *gin.Context) {
...
+// @tags provider
+// @accept json
+// @produce json
+// @param provider body Provider _ "provider object"
+// @success 200 {object} Provider
+// @failure 400 {object} problem.Response "Bad request"
+// @failure 401 {object} problem.Response "Unauthorized or missing jwt token"
+// @failure 502 {object} problem.Response "Database is unreachable"
+// @router /provider [put]
+func (m ProviderModule) putProviderHandler(c *gin.Context) {
```
jilleJr, line 201: Forgotten to add this method to the `ProviderModule.Register` method

jilleJr (changes requested):
I forgot about this detail :) Lucky that your other PR (#26) brought this up

provider.go
fredx30: Its in!
https://github.com/iver-wharf/wharf-api/blame/63c2c4986a42f58394f8db8729265fe5ba154a6d/provider.go#L42

jilleJr (approved):
LGTM!

iverestefans (approved):

PR #29, opened 2021-06-03
Expose bind address as env var BIND_ADDRESS

fredx30:
This allows setting the gin bind adress and port from an environment variable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 -->
 
 ## v4.1.0 (WIP)
+
 - Added environment var for setting bind address and port. (#29)
 
 - Fixed missing `failed` field from `main.TestsResults` in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 	https://changelog.md/
 -->
 
-## v4.0.1 (WIP)
+## v4.1.0 (WIP)
+- Added environment var for setting bind address and port. (#29)
 
 - Fixed missing `failed` field from `main.TestsResults` in
   `GET /build/{buildid}/tests-results`. (#25)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,7 +12,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 	https://changelog.md/
 -->
 
-## v4.0.1 (WIP)
+## v4.1.0 (WIP)
+- Added environment var for setting bind address and port. (#29)
```
jilleJr, line 16: nit: missing empty line after header

```suggestion

- Added environment var for setting bind address and port. (#29)
```

main.go
```diff
@@ -123,7 +123,12 @@ func main() {
 	api.GET("/version", getVersionHandler)
 	api.GET("/swagger/*any", ginSwagger.WrapHandler(swaggerFiles.Handler))
 
-	_ = r.Run()
+	var bindAddress string
+	var bindAddressDefined bool
+	if bindAddress, bindAddressDefined = os.LookupEnv("BIND_ADDRESS"); !bindAddressDefined {
+		bindAddress = "0.0.0.0:8080"
+	}
+	_ = r.Run(bindAddress)
```
jilleJr, line 131: Gin does read port by default from environment variable `PORT` https://github.com/gin-gonic/gin/blob/34ce2104cad324f444943c528746bf6d23643cd3/utils.go#L139-L153

(it's not well documented in though)

So we're only gaining address here. But it's a nice-to-have feature for those that need it though.

jilleJr (approved):
A nit on the CHANGELOG

main.go
jilleJr, line 131: Just want to also apply the same comment here as over at 
https://github.com/iver-wharf/wharf-provider-azuredevops/pull/11#discussion_r645333238 : 

I think it'd be better if this was extracted to its own function.

```go
func getBindAddress() string {
	bindAddress, ok := os.LookupEnv("BIND_ADDRESS")
	if !ok || bindAddress == "" {
		return "0.0.0.0:8080"
	}
	return bindAddress
}
```

jilleJr (approved):

main.go
fredx30: Cool didnt know about the port binding. This will let me bind the adress to the docker vlan though which is a plus.

fredx30: Done.

Pikabanga (approved):

PR #33, opened 2021-06-14
Changed references to wharf-core pkg/problem and pkg/ginutil

### LINKED ISSUE #32: Use Problem from wharf-core
Problem type was moved to wharf-core in https://github.com/iver-wharf/wharf-core/pull/9 so this repo needs to update to reference that one instead.

Alexamakans:
Closes #32


fredx30 (approved):
Nice. I checked 2 of the codeql warnings and they seem like they are the false positives i have been dealing with before.

jilleJr (approved):
LGTM!

PR #43, opened 2021-08-19
Reworked handling of test results

### LINKED ISSUE #11: Test Results Summary
Right now the test results are taken for each build separately and calculated each time by parsing artifact files.

A new table needs to be created to keep the test results summary. That table should have relations N:1 to build table and 1:1 relation to an artifact table.

Table should contain the number of tests run, passed, failed and skipped.

The values to this table should be calculated once when the artifact is uploaded into the backend.

Create a separate method to upload test results apart from artifacts

Modify the get method for builds to include the total test numbers in the build entity. That number should be a sum of all entities in the test result summary table related to that build.

The get test results method should be modified to return the list of results + a link to an artifact that those results originated from (artifact Id should be enough)

Change the frontend to use the new way of getting test results instead of making those additional calls per build.

image

-----

- [ ] Make the database changes
- [ ] Fill out the data from existing artifacts
- [ ] Create an Upload TestResult endpoint. This will parse the test result files (UTs)
- [ ] Modify the existing wharf-ci configuration to use new endpoint (in our other internal projects that use Wharf with test results)
- [ ] Extend the Build entity with TestResultSummary
- [ ] Use new data from Build entity on the Frontend
- [ ] Remove /testResults endpoint

### LINKED ISSUE #17: Test Results Details
In comparison to #11 , this issue is for full test results. Meaning each single test should be stored with its state and messages.

Should be populated when a test result is uploaded

Need endpoints to be able to fetch these.

Could probably settle for only storing non-successful tests

### LINKED ISSUE #66: Rework of test result handling
> Based on [RFC-0014 iver-wharf/rfcs#14](https://github.com/iver-wharf/rfcs/pull/14)

Changing handling of test results in wharf-api.

Parse the .trx files once and insert into database instead of parsing each time it's requested.
Also gives the possibility to get more detailed summaries, as well as more detailed individual test results.

Please see the RFC for further details: [RFC-0014](https://github.com/iver-wharf/rfcs/pull/14)

Alexamakans:

## Summary

Implementation of: https://github.com/iver-wharf/rfcs/blob/master/_published/0014-test-results-handling.md
Closes #11, closes #17, closes #66

Added 2 new database models

- `TestResultDetail`, represents a test function.
  - Name, outcome, start time, end time, message of a test function.
- `TestResultSummary`, represents a UT.
  - Filename (.trx file), number of failed/skipped/passed tests in it, as well as the total amount of test functions.

Added 3 new response models

- `TestResultDetail`, same as above.
- `TestResultSummary`, same as above.
- `TestResultListSummary`, represents a test run.
  - Sums up failed/skipped/passed tests of the summaries of a test run, and the total number of tests that ran.

Added 6 new endpoints
- `POST /build/{buildid}/test-result`
  - Takes a TRX file, inserts it as an `Artifact` to the `artifacts` table.
  - Parses the files into:
    - `TestResultDetail`, one or more per file.
    - `TestResultSummary`, one per file.
    - These are inserted into the `test_result_detail` and `test_result_summary` tables respectively.
  - The response is an array of the created artifact's filenames and artifact IDs. 
- `GET /build/{buildid}/test-result/detail`
  - Gets all the test result details for the specified build.
- `GET /build/{buildid}/test-result/summary/{artifactId}/detail`
  - Gets all the test result details for the specified test.
- `GET /build/{buildid}/test-result/summary`
  - Gets all the test result summaries for the specified build.
- `GET /build/{buildid}/test-result/summary/{artifactId}`
  - Gets the test result summary for the specified test.
- `GET /build/{buildid}/test-result/list-summary`
  - Gets the test result list summary of all tests for the specified build.

Deprecated `GET /build/{buildid}/tests-results`
- Use `GET /build/{buildid}/test-result/list-summary` instead. The response data is slightly different; it has additional properties, and does not have a `status` property.

## Motivation

Without this we were unable to retrieve any pertinent information regarding test results, besides seeing if _all_ tests passed or not.

With this change we can make it so we can see which tests failed, and the message it failed with over in wharf-web. In addition to this, we can also make it so it shows how many tests failed/skipped/passed in the builds list, without having to go to the specific build's detail page.

This should help with debugging.

## Notes

We also need to change the build step (phrasing? Not the building of Wharf itself, but when using Wharf to run something) so that it uploads the test result files to the new endpoint `POST /build/{buildid}/test-result`.

## Glossary

I'm not sure what words to use because they all seem really similar to me, but this is how I've tried to use them in the PR.
Order: small -> big
`test function`: A single test function. = `TestResultDetail`
`UT`: A single test file, collection of test functions. = `TestResultSummary`
`test run`: A collection of UTs, multiple files possible. = `TestResultListSummary`


jilleJr:
In your Summary you need to preface each issue reference with the keyword "closes" to automatically close multiple issues

https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#closing-multiple-issues

artifact.go
```diff
@@ -8,51 +8,42 @@ import (
...
-	Failed int        `json:"failed"`
-	Status TestStatus `json:"status" enums:"Success,Failed,No tests"`
+// SummaryOfTestResultSummaries contains data about several test result files.
+type SummaryOfTestResultSummaries struct {
+	BuildID   uint                `json:"buildId"`
+	Total     uint                `json:"total"`
+	Failed    uint                `json:"failed"`
+	Passed    uint                `json:"passed"`
+	Skipped   uint                `json:"skipped"`
+	Summaries []TestResultSummary `json:"summaries"`
```
jilleJr, line 34: That name though...

Could perhaps be enough with only 1 "Summary" in the name.

What about this?:

```go
// TestResultListSummary contains data about several test result files.
type TestResultListSummary struct {
	BuildID uint                `json:"buildId"`
	Total   uint                `json:"total"`
	Failed  uint                `json:"failed"`
	Passed  uint                `json:"passed"`
	Skipped uint                `json:"skipped"`
	Results []TestResultSummary `json:"results"`
}
```

or maybe:

```go
// BuildTestResultSummary contains summary of tests for a given build.
type BuildTestResultSummary struct {
	BuildID uint                `json:"buildId"`
	Total   uint                `json:"total"`
	Failed  uint                `json:"failed"`
	Passed  uint                `json:"passed"`
	Skipped uint                `json:"skipped"`
	Results []TestResultSummary `json:"results"`
}
```

artifact.go
```diff
@@ -163,50 +151,311 @@ func (m ArtifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+
+	for _, artifact := range artifacts {
+		detail, summary, err := parseAsTRX(artifact.Data)
+		if err != nil {
+			log.Warn().
+				WithError(err).
+				WithString("filename", artifact.Name).
+				WithUint("build", buildID).
+				WithUint("artifact", artifact.ArtifactID).
+				Message("Failed to unmarshal; invalid JSON format")
```
jilleJr, line 279: This message seems off. How do you know it's because of invalid JSON format?

artifact.go
```diff
@@ -163,50 +151,311 @@ func (m ArtifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+			continue
+		}
+
+		for _, d := range detail {
```
jilleJr, line 283: The variable `detail` is singular, yet it seems to contain more than one detail.

artifact.go
```diff
@@ -163,50 +151,311 @@ func (m ArtifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+
+		summaries = append(summaries, summary)
+		details = append(details, detail...)
+	}
+
+	summaryOfSummaries := SummaryOfTestResultSummaries{
+		Failed:    0,
+		Passed:    0,
+		Skipped:   0,
+		Total:     0,
```
jilleJr, line 300: These defaults to zero. Think it will be equally readable if you just omitted these declarations

artifact.go
```diff
@@ -163,50 +151,311 @@ func (m ArtifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+		Summaries: make([]TestResultSummary, 0, len(summaries)),
```
jilleJr, line 301: Could you not assign the array of summaries here directly?

artifact.go
```diff
@@ -163,50 +151,311 @@ func (m ArtifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+	for _, artifact := range artifacts {
+		detail, summary, err := parseAsTRX(artifact.Data)
+		if err != nil {
+			log.Warn().
+				WithError(err).
+				WithString("filename", artifact.Name).
+				WithUint("build", buildID).
+				WithUint("artifact", artifact.ArtifactID).
+				Message("Failed to unmarshal; invalid JSON format")
+			continue
```
jilleJr, line 280: I think an invalidly formatted TRX file should not be silently ignored

artifact.go
```diff
@@ -163,50 +151,311 @@ func (m ArtifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+}
+
+// parseMultipartFormData writes 400 "Bad request" problem.Response on failure.
+// Returns a slice of file pointers on success, or an empty slice if there were
+// none but parsing was successful.
+func parseMultipartFormData(c *gin.Context) ([]*file, bool) {
+	buildID, ok := parseBuildID(c)
+	if !ok {
+		return nil, false
+	}
```
jilleJr, line 449: Dependency on the `{buildid}` parameter is quite hidden in this function. I suggest to accept it through a parameter or remove it from the problem detail strings

artifact.go
```diff
@@ -216,39 +465,193 @@ func (m ArtifactModule) postBuildArtifactHandler(c *gin.Context) {
...
-				Data:     data,
-				Name:     k,
-				FileName: fhs[0].Filename,
-				BuildID:  buildID,
-			}
+			files = append(files, &file{
+				name:     k,
+				fileName: fhs[0].Filename,
+				data:     data,
+			})
```
jilleJr, line 483: IMO excess use of pointers. Suggest making the `file.data` field an enforced pointer instead, so you don't have to spray pointer dereferencing and such everywhere

jilleJr (changes requested):

artifact.go
Alexamakans: From some googling it seems like the slice's underlying array isn't copied when passing the slice by value. With this in mind I have changed it so that all the things are no longer pointers.

Not sure if that's the way to go, but the code looks cleaner now.

I also made sure to initialize the slices with a length to avoid unnecessary allocations.

jilleJr: I always get a little confused by Go's difference between slice vs array. But if I understand it correctly, you're always working with slices unless you specify the size inside the brackets. And only if you specify the brackets, then the copy operation is expensive.

```go
var myArray = [3]int{1, 2, 3}

var mySlice = []int{1, 2, 3}
var myOtherSlice = make([]int, 3, 3)

var myArrayCopied = myArray // expensive op on big arrays
var myArrayAsSlice = myArray[:] // cheap op, just taking a ref

type myType struct {
    sliceField []int
    arrayField [3]int
}

ex1 := myType{sliceField: myArray}        // ok
ex2 := myType{sliceField: mySlice}        // ok
ex3 := myType{sliceField: make([]int, 3)} // ok

ex4 := myType{arrayField: myArray}        // ok, but expensive op on big arrays
ex5 := myType{arrayField: mySlice}        // compile error
ex6 := myType{arrayField: make([]int, 3)} // compile error
}
```

As you say, bottom line is that we're only working with slices anyway, and I think trying to use arrays here would be overkill when slices clearly suffices, even if we force the GC to deal with them as there's not 10 000 items in them

artifact.go
```diff
@@ -203,67 +213,423 @@ func (m artifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+	}
+
+	artifactID, ok := ginutil.ParseParamUint(c, artifactIDParamName)
+	if !ok {
+		return
+	}
+
+	details := []TestResultDetail{}
+	err := m.Database.
+		Where(&TestResultDetail{BuildID: uint(buildID), ArtifactID: artifactID}).
```
jilleJr, line 388: Is this cast really needed?

```suggestion
		Where(&TestResultDetail{BuildID: buildID, ArtifactID: artifactID}).
```

artifact.go
```diff
@@ -203,67 +213,423 @@ func (m artifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+func (m artifactModule) getBuildTestResultsSummaryHandler(c *gin.Context) {
+	buildID, ok := ginutil.ParseParamUint(c, buildIDParamName)
+	if !ok {
+		return
+	}
+
+	summaries := []TestResultSummary{}
+	err := m.Database.
+		Preload("Artifact").
+		Where(&TestResultSummary{BuildID: uint(buildID)}).
```
jilleJr, line 419: Is this cast really needed?

```suggestion
		Where(&TestResultSummary{BuildID: buildID}).
```

_hint: Running `npm run lint-go` will probably reveal these unneeded casts, as well as other minor scripting errors_

artifact.go
```diff
@@ -203,67 +213,423 @@ func (m artifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+		summaryList.Skipped += v.Skipped
+	}
+
+	summaryList.Total = summaryList.Failed + summaryList.Passed + summaryList.Skipped
+
+	c.JSON(http.StatusOK, summaryList)
+}
+
+func parseMultipartFormData(c *gin.Context, buildID uint) ([]file, bool) {
+	files := make([]file, 0)
```
jilleJr, line 446: Suggest moving this declaration to just above the `for` loop where it's used. And you don't need to `make` up a new slice, as nil slices and empty slices are treated equally.

```go
append([]int{}, 1) // => []int{ 1 }
append(make([]int), 2) // => []int{ 2 }
append([]int(nil), 3) // => []int{ 3 }

len([]int{}) // => 0
len(make([]int)) // => 0
len([]int(nil)) // => 0

// the following are semantically identical:
var files []file
var files []file = nil
var files []file = []file(nil)
var files        = []file(nil)
```

```suggestion
	var files []file
```

artifact.go
```diff
@@ -203,67 +213,423 @@ func (m artifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
 		if fhs := form.File[k]; len(fhs) > 0 {
-			f, err := fhs[0].Open()
-			if err != nil {
-				ginutil.WriteMultipartFormReadError(c, err, fmt.Sprintf(
-					"Failed with starting to read file content from multipart form request body when uploading new artifact for build with ID %d.",
-					buildID))
-				return
+			data, ok := readMultipartFileData(c, buildID, fhs[0])
+			if !ok {
+				return files, false
```
jilleJr, line 460: Vote fot returning nil here as well, as it failed halfway. So it's consistent with the returns for the other fail states, and that we don't expect to use the return value anyway.

```suggestion
				return nil, false
```

artifact.go
```diff
@@ -203,67 +213,423 @@ func (m artifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
+type testRun struct {
+	XMLName       xml.Name      `xml:"TestRun"`
+	Results       results       `xml:"Results"`
+	ResultSummary resultSummary `xml:"ResultSummary"`
+}
+
+type results struct {
+	XMLName         xml.Name         `xml:"Results"`
+	UnitTestResults []unitTestResult `xml:"UnitTestResult"`
+}
```
jilleJr, line 507: Some of these types has quite generic names. Especially the `results`, `output`, `message`, `stackTrace`, and `counters` ones. Could you prefix them with something like `xml` or `trx` to make it clearer where they're used, as well as to not produce unexpected behaviour from name collisions? Alternatively move these types to their own package, such as `/pkg/testresults/fbxtests` or something, together with the parsing

Also, you can nest anonymous structs in other structs. Maybe it's something you want to apply here?

```go
type trxTestRun struct {
	XMLName xml.Name `xml:"TestRun"`
	Results struct {
		XMLName         xml.Name `xml:"Results"`
		UnitTestResults []struct {
			XMLName   xml.Name `xml:"UnitTestResult"`
			TestName  string   `xml:"testName,attr"`
			Duration  string   `xml:"duration,attr"`
			StartTime string   `xml:"startTime,attr"`
			EndTime   string   `xml:"endTime,attr"`
			Outcome   string   `xml:"outcome,attr"`
			Output    output   `xml:"Output"`
		} `xml:"UnitTestResult"`
	} `xml:"Results"`
	ResultSummary resultSummary `xml:"ResultSummary"`
}
```

Doing so makes it extreamly difficult to create the subtypes separately without having to redeclare the entire type (such as in function parameters), but when you only work with the full object then this works really niceley

database_models.go
```diff
@@ -160,9 +161,49 @@ const (
 // the file name and which build it belongs to.
 type Artifact struct {
 	ArtifactID uint   `gorm:"primaryKey" json:"artifactId"`
-	BuildID    uint   `gorm:"not null;index:param_idx_build_id" json:"buildId"`
+	BuildID    uint   `gorm:"not null;index:artifact_idx_build_id" json:"buildId"`
```
jilleJr, line 164: Don't know how this has worked before, how GORM has resolved this name conflict, but if GORM had set up the index named `param_idx_build_id` then I think you need to add a migration step to remove the said index from the `artifacts` table. I've not double-checked, but is the `param_idx_build_id` index only applied to one of the tables or both, and will this change make any duplicates or will it resolve itself by GORM's automatic migrations?

database_models.go
```diff
@@ -160,9 +161,49 @@ const (
...
 	FileName   string `gorm:"nullable" json:"fileName"`
 	Data       []byte `gorm:"nullable" json:"-"`
 }
+
+// TestResultSummary contains data about a single test result file.
+type TestResultSummary struct {
+	TestResultSummaryID uint      `gorm:"primaryKey" json:"testResultSummaryId"`
+	FileName            string    `gorm:"nullable" json:"fileName"`
+	ArtifactID          uint      `gorm:"not null;index:testresultsummary_idx_artifact_id" json:"artifactId"`
+	Artifact            *Artifact `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
```
jilleJr, line 176: I think deleting the artifact should actually set this to null on the testresultsummary instead of locking this dependency. Thoughts? Otherwise I think a value of CASCADE would be more fitting here

database_models.go
```diff
@@ -160,9 +161,49 @@ const (
...
+	BuildID             uint      `gorm:"not null;index:testresultsummary_idx_build_id" json:"buildId"`
+	Build               *Build    `gorm:"foreignKey:BuildID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
```
jilleJr, line 178: Definetely think these OnUpdate/OnDelete should be CASCADE and not RESTRICT to make our lives easier. It leaves more to the database to deal with and less explicit code from us, but I think that abusing CASCADE is more of an obvious DB design than having RESTRICT everywhere

database_models.go
```diff
@@ -160,9 +161,49 @@ const (
...
+type TestResultDetail struct {
+	TestResultDetailID uint             `gorm:"primaryKey" json:"testResultDetailId"`
+	ArtifactID         uint             `gorm:"not null;index:testresultdetail_idx_artifact_id" json:"artifactId"`
+	Artifact           *Artifact        `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+	BuildID            uint             `gorm:"not null;index:testresultdetail_idx_build_id" json:"buildId"`
+	Build              *Build           `gorm:"foreignKey:BuildID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+	Name               string           `gorm:"not null" json:"name"`
+	Message            null.String      `gorm:"nullable" json:"message" swaggertype:"string"`
+	StartedOn          *time.Time       `gorm:"nullable;default:NULL;" json:"startedOn" format:"date-time"`
+	CompletedOn        *time.Time       `gorm:"nullable;default:NULL;" json:"completedOn" format:"date-time"`
```
jilleJr, line 207: Suggest using `null.Time` instead

```suggestion
	StartedOn          null.Time        `gorm:"nullable;default:NULL;" json:"startedOn" format:"date-time"`
	CompletedOn        null.Time        `gorm:"nullable;default:NULL;" json:"completedOn" format:"date-time"`

```

artifact.go
```diff
@@ -203,67 +213,423 @@ func (m artifactModule) getBuildTestsResultsHandler(c *gin.Context) {
...
-				WithUint("artifact", artifact.ArtifactID).
-				Message("File saved as artifact")
+	f, err := fh.Open()
+	if err != nil {
+		ginutil.WriteMultipartFormReadError(c, err,
+			fmt.Sprintf("Failed with starting to read file content from multipart form request body when"+
+				" uploading new artifact for build with ID %d.", buildID))
+		return nil, false
+	}
+	defer f.Close()
```
jilleJr, line 486: Can actually also fail during closing. Don't know the best way to handle that, but I think we should at least log it

```suggestion
	defer func() {
		if err := f.Close(); err != nil {
			log.Error().
				WithError(err).
				Message("Failed to close multipart form request body file handle.")
		}
	}
```

artifact.go
Alexamakans: Seems in line with what I read and understood 👍

artifact.go
Alexamakans: Decided to go with `TestResultListSummary`.

CHANGELOG.md
```diff
@@ -48,6 +48,30 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 - Changed version of `github.com/swaggo/swag/cmd/swag` from v1.7.0 to v1.7.1 in
```
jilleJr, line 48: A note on the glossary in the PR description:

> - test function: A single test function. = TestResultDetail
> - UT: A single test file, collection of test functions. = TestResultSummary
> - test run: A collection of UTs, multiple files possible. = SummaryOfTestResultSummaries

For one thing they are not guaranteed to be unit tests. Even TRX files can include integration tests, benchmark tests, end-to-end tests, et.al.

And a single uploaded test file/artifact with test results is more aligned with a "test run" and not a collection of files. Altough that's mostly true for our dotnet samples, but not sure if other test libs create multiple files for different subprojects, category, or whatever.

This is pretty late in the review process, so it's fully OK to ignore this suggestion, but maybe something like this:

> - test result: A single test = TestResult
> - test results file: A single test file, collection of tests. = TestResultSummary
> - build test results: A summary of test results for a given build, multiple files possible. = BuildTestResultSummary

artifact.go
```diff
@@ -215,55 +218,413 @@ func (m artifactModule) postBuildArtifactHandler(c *gin.Context) {
...
+// @router /build/{buildid}/test-result-details [get]
+func (m artifactModule) getBuildAllTestResultDetailsHandler(c *gin.Context) {
+	buildID, ok := ginutil.ParseParamUint(c, "buildid")
+	if !ok {
+		return
+	}
+
+	details := []TestResultDetail{}
+	err := m.Database.
+		Where(&TestResultDetail{BuildID: uint(buildID)}).
```
jilleJr, line 349: That cast seems to be unneeded.

```suggestion
		Where(&TestResultDetail{BuildID: buildID}).
```

artifact.go
```diff
@@ -215,55 +218,413 @@ func (m artifactModule) postBuildArtifactHandler(c *gin.Context) {
...
+	for _, artifact := range artifacts {
+		summary, details, err := getTestSummaryAndDetails(artifact.Data, artifact.ArtifactID, buildID)
+		if err != nil {
+			log.Warn().
+				WithError(err).
+				WithString("filename", artifact.Name).
+				WithUint("build", buildID).
+				WithUint("artifact", artifact.ArtifactID).
+				Message("Failed to unmarshal; invalid TRX/XML format, or the structure" +
+					" receiving the data is malformed.")
```
jilleJr, line 276: What do you mean with this message "the structure receiving the data is malformed"?

I think it's better to only state what has gone wrong, especially in log messages, and let the reader figure out what the cause it.

Ex:

```suggestion
				Message("Failed to unmarshal TRX/XML test results artifact.")
```

artifact.go
```diff
@@ -215,55 +218,413 @@ func (m artifactModule) postBuildArtifactHandler(c *gin.Context) {
...
+
+			ginutil.WriteProblemError(c, err,
+				problem.Response{
+					Type:   "/prob/unexpected-response-format",
+					Status: 502,
+					Title:  "Unexpected response format.",
+					Detail: fmt.Sprintf(
+						"Failed parsing test result ID %d, for build with ID %d in"+
+							" database. Expected TRX/XML format. Could also be due to an"+
+							" internal bug.", summary.ArtifactID, buildID),
```
jilleJr, line 286: This error message projects a lack of confidence :p 

Suggest eluding that "could be due to an internal bug" part. There's always the possibility of a scripting error, but I don't think we should suggest that in the error messages.

Everyone can write bugs, but we should not expect our code to have bugs. If that's the case, then we need to fix that. *(While there are parts of Wharf I expect has a lot of bugs in, but that's nothing we should aim for)*

```suggestion
							" database. Expected TRX/XML format.",
							summary.ArtifactID, buildID),
```

jilleJr (changes requested):

artifact.go
```diff
@@ -267,24 +268,37 @@ func (m artifactModule) postTestResultDataHandler(c *gin.Context) {
...
+					problem.Response{
+						Type:   "/prob/bad-code",
+						Status: http.StatusInternalServerError,
+						Title:  "Bad code.",
+						Detail: fmt.Sprintf(
+							"Failed parsing test result ID %d, for build with ID %d in database. The structure"+
+								" used to unmarshall the data in the wharf-api might be malformed. This really shouldn't happen.",
+							summary.ArtifactID, buildID),
+					})
+			}
```
Alexamakans: I checked the source for `xml.Unmarshal` to see what errors were returned, and came up with this.
It feels a bit massive, and not sure about the choice for the second problem.

Thoughts?

Alexamakans: Saw https://github.com/iver-wharf/wharf-api/pull/43#discussion_r695437525 just now.

With the split of plausible error vs. implausible error (in production), I feel that the `Bad code.` and `This really shouldn't happen.` messages are warranted, and makes it clear to both the users and developers that it is something that requires checking the code.

It's likely all errors that would go to the second branch would be found with unit test for that function, allowing us to "ignore" handling those as they wouldn't happen.

jilleJr: I'd suggest to change to only check `err != nil` and then return a problem for "Failed to parse TRX test results file" and then let the error itself explain the nitty-gritty reason for it.

You can still have the problem detail be something like it is now with "the data might be malformed" or "Invalid TRX/XML file format".

The problem types here are kind of a stretch IMO. Perhaps instead have `Type: "/prob/api/test-results-parse"` or something, where the docs could suggest that you uploaded a file with the wrong format or that you selected the wrong format in the query parameters or something.

artifact.go
```diff
@@ -267,24 +268,37 @@ func (m artifactModule) postTestResultDataHandler(c *gin.Context) {
...
-					Type:   "/prob/unexpected-response-format",
-					Status: 502,
-					Title:  "Unexpected response format.",
-					Detail: fmt.Sprintf(
-						"Failed parsing test result ID %d, for build with ID %d in"+
-							" database. Expected TRX/XML format. Could also be due to an"+
-							" internal bug.", summary.ArtifactID, buildID),
-				})
+				WithUint("artifact", artifact.ArtifactID)
+			if syntaxErr := err.(*xml.SyntaxError); syntaxErr != nil || strings.HasPrefix(err.Error(), "xml:") {
```
jilleJr, line 276: This cast will fail if the `syntaxErr` is not of the type `*xml.SyntaxError`.

A safer bet when casting is to use the second return value from the cast, like so:

```go
var err error = errors.New("foo")

syntaxErr := err.(*xml.SyntaxError) // panic

syntaxErr, ok := err.(*xml.SyntaxError) // ok = false
```

But [since Go v1.13 you should instead use `errors.As` and `errors.Is` for these cases](https://go.dev/blog/go1.13-errors), as they handle nested errors

```suggestion
			var syntaxErr *xml.SyntaxError
			if errors.As(err, &syntaxErr) {
```

jilleJr (changes requested):

artifact.go
Alexamakans: Looking closer at the source, the implementation of `Close()` for the file returned is just returning nil, so we can remove it completely. Although this is only true when the `multipart.FileHeader`'s content property is non-nil. I made sure to check that that was the case here.

jilleJr: While this is true now, it could change later on. To follow "best coding practices", you should always at least close a file handle when you're done with it, even if the underlying implementation does nothing in this version.

Suggest at least reverting commit 64181eb

database_models.go
Alexamakans: I agree.

Added a short description in b8dd4f2 explaining the reasoning behind it, or at least my reasoning.
Took the liberty to change it for `TestResultDetail` as well.

test_result.go
```diff
@@ -0,0 +1,373 @@
...
+				XMLName   xml.Name `xml:"Output"`
+				ErrorInfo struct {
+					XMLName xml.Name `xml:"ErrorInfo"`
+					Message struct {
+						InnerXML string `xml:",innerxml"`
+					} `xml:"Message"`
+					StackTrace struct {
+						InnerXML string `xml:",innerxml"`
+					} `xml:"StackTrace"`
+				} `xml:"ErrorInfo"`
```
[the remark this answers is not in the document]
Alexamakans: Didn't fully understand how it worked. Seems only one is needed.
f6b1bb4

test_result.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	listSummary := TestResultListSummary{
+		BuildID: buildID}
+
+	for _, v := range summaries {
+		listSummary.Failed += v.Failed
+		listSummary.Passed += v.Passed
+		listSummary.Skipped += v.Skipped
+	}
```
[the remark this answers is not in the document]
Alexamakans: Seem to work nicely 👍
f6b1bb4

Alexamakans:
> This looks good to me.
> Something to consider for future PRs though- I got in a bit late- after 2 rounds of lots of comments and read through them all. It would be easier to get a historical overview over what has done if the comments that requested change were answered somehow.
> If the change request has an indication of how it should be done, and you implemented the change in the suggested way one way to convey this in the PR is to react with a thumb up emoj to the comment before resolving. If the comment is resolved in a way other than the suggested way a comment indicating a summary of the actual change is needed. To close the feedback loop in this case the reviewer would then have to react positively to that comment or continue the thread with more feedback.

I like that idea. I will try to keep that in mind in the future.

test_result.go
```diff
@@ -0,0 +1,373 @@
...
+		listSummary.Skipped += summary.Skipped
+
+		summary.FileName = artifact.FileName
+		summaries = append(summaries, summary)
+		lotsOfDetails = append(lotsOfDetails, details...)
+	}
+
+	listSummary.Total = listSummary.Failed + listSummary.Passed + listSummary.Skipped
+
+	if err := m.Database.CreateInBatches(summaries, 10).Error; err != nil {
```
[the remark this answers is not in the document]
Alexamakans: Opting to leave this as-is for now.

artifact.go
```diff
@@ -149,11 +139,38 @@ func (m artifactModule) getBuildArtifactHandler(c *gin.Context) {
...
+		return
+	}
+
+	c.Status(http.StatusCreated)
+}
+
 // getBuildTestsResultsHandler godoc
-// @summary Get build tests results from .trx files
+// @deprecated
+// @summary Get build tests results from .trx files. Deprecated, /build/{buildid}/test-results-summary should be used instead.
```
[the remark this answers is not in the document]
Alexamakans: This is the now deprecated endpoint that specifically looked for .trx files, as can be seen [here](https://github.com/iver-wharf/wharf-api/pull/43/files?file-filters%5B%5D=.go&file-filters%5B%5D=.mod&file-filters%5B%5D=.sum#diff-69f26222afc2d6b0701abcbf24bc8d3784ba2d30189c65c9d3c39c1d1cbc14e5R179)

internal/ctxparser/ctxparser.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// ParamBuildIDAndFiles is a helper function that parses the param
+// buildid and multipart form data files from the gin.Context.
+func ParamBuildIDAndFiles(c *gin.Context) (buildID uint, files []File, ok bool) {
+	if buildID, ok = ginutil.ParseParamUint(c, "buildid"); ok {
+		files, ok = parseMultipartFormData(c, buildID)
+	}
+	return
+}
```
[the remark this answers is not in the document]
Alexamakans: Opted to revert the parsing of `artifactId` and `buildid` to using `ginutil.ParseParamUint`.
Exported `ctxparser.ParseMultipartFormData`.

Alexamakans:
Converted to draft because it depends on https://github.com/iver-wharf/wharf-api/pull/54

database_models.go
```diff
@@ -112,18 +112,19 @@ var buildJSONToColumns = map[string]string{
...
+	Project             *Project            `gorm:"foreignKey:ProjectID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+	ScheduledOn         null.Time           `gorm:"nullable;default:NULL" json:"scheduledOn" format:"date-time"`
+	StartedOn           null.Time           `gorm:"nullable;default:NULL" json:"startedOn" format:"date-time"`
+	CompletedOn         null.Time           `gorm:"nullable;default:NULL" json:"finishedOn" format:"date-time"`
+	GitBranch           string              `gorm:"size:300;default:'';not null" json:"gitBranch"`
+	Environment         null.String         `gorm:"nullable;size:40" json:"environment" swaggertype:"string"`
+	Stage               string              `gorm:"size:40;default:'';not null" json:"stage"`
+	Params              []BuildParam        `gorm:"foreignKey:BuildID" json:"params"`
+	IsInvalid           bool                `gorm:"not null;default:false" json:"isInvalid"`
+	TestResultSummaries []TestResultSummary `gorm:"foreignKey:BuildID" json:"testResultSummaries"`
```
fredx30, line 127: All spacing here is changed resulting in a 100% diff for these lines. Ideally the spacing change would be in different PR to make this more reviewable.

There is a hidden `null.Time` -> `*time.Time ` in this change- Intentional?

fredx30 (approved):

database_models.go
Alexamakans: The change from `*time.Time` to using `null.Time` was intentional, more of a drive-by change than required for the implementation.


I agree that it really makes it harder to see what's going on. I think it would cause more trouble if we tried to avoid it though, as it gets formatted like that when you run `goimports -w .`, and the pros of formatting while coding outweigh the cons in this case IMO. Especially if it would mean an extra "formatting" PR after basically every PR that touches a struct.

This is mainly an issue for structs I believe, and big ones at that, so I think while it's not great when it's something like this, it shouldn't happen very often. One solution/workaround could be to use `git diff --word-diff` if you're checking locally, or https://diffonline.net/ [4-color (light/dark red/green) highlighting] or https://text-compare.com/ [1-color highlighting].

Unless there is a way to change the GitHub diff settings somehow, which would be great.

jilleJr: There is a way to hide changes if it's only whitespace changes on the given line.

Screenshot from 2021-09-15 08-01-02

Before:

Screenshot from 2021-09-15 08-02-18

After:

Screenshot from 2021-09-15 08-00-27

Alexamakans: Awesome! 👍

test_result.go
```diff
@@ -0,0 +1,405 @@
...
+)
+
+type buildTestResultModule struct {
+	Database *gorm.DB
+}
+
+type artifactListItem struct {
+	FileName   string `json:"fileName"`
+	ArtifactID uint   `json:"artifactId"`
+}
```
jilleJr, line 23: nit: Because of it's usage, I suggest a name such as `ArtifactMetadata`.

test_result.go
```diff
@@ -0,0 +1,405 @@
...
+
+// TestResultListSummary contains data about several test result files.
+type TestResultListSummary struct {
+	BuildID uint `json:"buildId"`
+	Total   uint `json:"total"`
+	Failed  uint `json:"failed"`
+	Passed  uint `json:"passed"`
+	Skipped uint `json:"skipped"`
+}
```
jilleJr, line 32: Most of your types here are exported, but not `artifactListItem`. Suggest exporting it as well (capitalizing it) for consistency.

test_result.go
```diff
@@ -0,0 +1,405 @@
...
+		testResult.GET("/list-summary", m.getBuildTestResultListSummaryHandler)
+	}
+}
+
+// postBuildTestResultDataHandler godoc
+// @summary Post test result data
+// @tags test-result
+// @accept multipart/form-data
+// @param buildid path int true "Build ID"
+// @param file formData file true "Test result file"
```
jilleJr, line 54: Here you only specify the form field `file`, while the `ctxparser.ParseMultipartFormData` reads from all files.

According to https://swagger.io/docs/specification/describing-request-body/file-upload/ to allow multiple files to be uploaded, you specify it as an array and then reuse the same form field name but upload multiple files.

Not sure if Swaggo supports it, but maybe the syntax is like so?

```go
// @param files formData []file true "Test result files"
```

And then I suggest you update ParseMultipartFormData to only parse a given form field's files.

test_result.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	listSummary.Total = listSummary.Failed + listSummary.Passed + listSummary.Skipped
+
+	c.JSON(http.StatusOK, listSummary)
+}
+
+type xmlInnerString struct {
+	InnerXML string `xml:",innerxml"`
+}
```
jilleJr, line 303: Why not embed this type as well inside `trxTestRun`?

Alexamakans: I decided to keep it outside because it's used twice.

test_result.go
jilleJr: Can't come up with any security issues.

As the swaggo issue is only for the documentation and not for the actual behavior, I suggest changing `ParseMultipartFormData` to only read from a given form field name (as I mentioned earlier), but then read all files from that field instead of only the first one in the array as it's doing right now.

Then once swaggo supports OpenAPI v3 we can update the field to expect an array of files instead of a single file.

As for the current usage in other projects: AFAIK it's one file per request.

Alexamakans: > As the swaggo issue is only for the documentation and not for the actual behavior, I suggest changing `ParseMultipartFormData` to only read from a given form field name (as I mentioned earlier), but then read all files from that field instead of only the first one in the array as it's doing right now.

Alright, will leave the methods as-is then, as there is the possibility of multiple files.

database_models.go
Alexamakans: @fredx30 Mentioning to make sure you see it too 👍🏻

test_result.go
Alexamakans: Done in 4a9871d

jilleJr (approved):

PR #44, opened 2021-08-19
Added /api/health and /api/ping endpoints

Alexamakans:

## Summary

Added `/api/health` and `/api/ping` endpoints, in preparation of removing the `/health` and `/` endpoints respectively.

## Motivation

The `/health` and `/` endpoints were not accessible through the Swagger API, as it tried to send requests to `/api/health` and `/api/ping` instead. This fixes that issue, and has all endpoints in the same group, i.e. `/api`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 - Added Makefile to simplify building and developing the project locally. (#41)
 
+- Fixed `/health` and `/ping` endpoints.
+
 ## v4.1.1 (2021-07-12)
 
 - Changed version of Docker base images:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 - Added Makefile to simplify building and developing the project locally. (#41)
 
+- Added new endpoints `/api/ping` and `/api/health`. (#44)
+
 - Deprecated `/` and `/health` endpoints, soon to be moved to `/api/ping` and
   `/api/health` respectively, so they are aligned with the current Swagger
   documentation. (#44)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,9 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 - Added Makefile to simplify building and developing the project locally. (#41)
 
-- Fixed `/health` and `/ping` endpoints.
+- Fixed `/health` and `/` endpoints to instead listen on `/api/health` and
+  `/api/ping`, respectively, so they are aligned with the current Swagger
+  documentation. (#44)
 
 ## v4.1.1 (2021-07-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 - Added Makefile to simplify building and developing the project locally. (#41)
 
-- Fixed `/health` and `/` endpoints to instead listen on `/api/health` and
-  `/api/ping`, respectively, so they are aligned with the current Swagger
+- Deprecated `/` and `/health` endpoints, soon to be moved to `/api/ping` and
+  `/api/health` respectively, so they are aligned with the current Swagger
   documentation. (#44)
 
 ## v4.1.1 (2021-07-12)
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/iver-wharf/wharf-api
 go 1.16
 
 require (
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
 	github.com/dustin/go-broadcast v0.0.0-20171205050544-f664265f5a66
 	github.com/ghodss/yaml v1.0.0
 	github.com/gin-contrib/cors v1.3.1
--- a/main.go
+++ b/main.go
@@ -116,8 +116,8 @@ func main() {
 		r.Use(cors.New(corsConfig))
 	}
 
-  healthModule{}.DeprecatedRegister(r)
-  healthModule{}.Register(r.Group("/api"))
+	healthModule{}.DeprecatedRegister(r)
+	healthModule{}.Register(r.Group("/api"))
 
 	setupBasicAuth(r, config)
 
--- a/main.go
+++ b/main.go
@@ -76,7 +76,7 @@ func main() {
 		log.Error().WithError(err).Message("Migration error")
 		os.Exit(3)
 	}
-  
+
 	mq, err := GetMQConnection(config.MQ)
 	if err != nil {
 		log.Error().WithError(err).Message("Message queue error.")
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 - Added Makefile to simplify building and developing the project locally. (#41)
 
+- Fixed `/health` and `/ping` endpoints.
```
jilleJr, line 28: What got fixed with the endpoints?

Maybe something like:

```suggestion
- Fixed `/health` and `/` endpoints to instead listen on `/api/health` and
  `/api/ping`, respectively, so they are aligned with the current Swagger
  documentation. (#44)
```

main.go
```diff
@@ -91,8 +91,6 @@ func main() {
 		r.Use(cors.New(corsConfig))
 	}
 
-	HealthModule{}.Register(r)
-
 	setupBasicAuth(r, config)
```
jilleJr, line 95: I think this was deliberately like this so that the basic auth was not applied to the HealthModule's endpoints. Try setting the auth environment variable and see if you can still access these endpoints anonymously, or if you need to revert the changes in this file.

If you want to make them require auth as well, I suggest you do that in a different PR as we need to make sure that our helm chart will properly provide Kubernetes with the right credentials when it's doing the liveness/readiness probes.

IMO leaving them as anonymous just simplifies so much.

---

One of the quirks with how we have it set up now is that you can only access the `/` and `/health` endpoints if you're inside the network, such as how Kubernetes is, while all users will not be given access as we have reverse proxies filtering out everything not starting with `/api/`. So we don't even need auth for those endpoints as they are not accessible to the "public" anyway.

Perhaps we could change the swagger documentation instead, and leave the endpoints as-is?

ping.go
```diff
@@ -6,9 +6,9 @@ import (
 
 type HealthModule struct{}
 
-func (m HealthModule) Register(e *gin.Engine) {
-	e.GET("/", m.ping)
-	e.GET("/health", m.health)
+func (m HealthModule) Register(g *gin.RouterGroup) {
+	g.GET("/ping", m.ping)
```
jilleJr, line 10: Still having an endpoint listening on GET requests on the root path is nice just so that it's easier to debug when setting things up. If you start up the wharf-api and then go to its IP and port via the web browser, having something respond instead of the 404 error from the browser is very helpful.

jilleJr (changes requested):
You can access the endpoints if you got access to the pod without the reverse proxy. Such as via `kubectl port-forward` or via http://localhost:5001/ if you're hosting via docker-compose. This is an advantage, though I can see how the swagger documentation inconsistency makes it confusing.

Screenshot from 2021-08-19 11-44-44

main.go
Alexamakans: A lot more behind this way of doing it than I expected.

It seems like changing the swagger documentation instead would be simpler and, dare I say, just as effective?
In that case I think that would be the best way to do it. I don't know in what way to go about changing the swagger docs though. If I get pointed in the right direction then I can do it, if nobody else wants to do it that is.

jilleJr: Hmm so if we change the swagger docs then you still wont be able to access it as the reverse proxy will point the requests elsewhere (probably towards the frontend). Changing the endpoints here does help with that.

What I propose at least is to keep the old endpoints for backward compatibility (look I spelled it right this time!) and move the endpoints as you have. So they're accessible from both places, and in next major release or something we can remove the old endpoints and perhaps replace the `/` endpoint with something else, like just a simple "hello from wharf-api" or something.

And then also add in the CHANGELOG.md a "Deprecated" note about the `/` and `/health` endpoints being moved soon to `/api/health` and `/api/ping`

Alexamakans: Pushed new commits.

> backward compatibility (look I spelled it right this time!)

Wooh!

ping.go
```diff
@@ -6,7 +6,12 @@ import (
 
 type HealthModule struct{}
 
-func (m HealthModule) Register(e *gin.Engine) {
+func (m HealthModule) Register(g *gin.RouterGroup) {
+	g.GET("/ping", m.ping)
+	g.GET("/health", m.health)
+}
+
+func (m HealthModule) DeprecatedRegister(e *gin.Engine) {
```
jilleJr, line 14: Would be nice with a comment here saying why it's deprecated and when it will potentially be removed. I feel that v5 is on the horizon, so maybe remove it in v6?

jilleJr (changes requested):

main.go
```diff
@@ -120,7 +120,8 @@ func main() {
 		BuildModule{Database: db, MessageQueue: mq},
 		TokenModule{Database: db},
 		BranchModule{Database: db},
-		ProviderModule{Database: db}}
+		ProviderModule{Database: db},
+		HealthModule{}}
```
jilleJr, line 124: These are still not anonymous. Suggest moving the deprecated registration to before the basic auth as well

Screenshot from 2021-08-23 09-13-09

ping.go
```diff
@@ -6,7 +6,14 @@ import (
...
 type HealthModule struct{}
 
-func (m HealthModule) Register(e *gin.Engine) {
+func (m HealthModule) Register(g *gin.RouterGroup) {
+	g.GET("/ping", m.ping)
+	g.GET("/health", m.health)
+}
+
+// Deprecated because they are not part of the /api group for endpoints.
+// Tentatively planned for complete removal in v6.
```
jilleJr, line 15: As Codacy/revive warns about here, doc comments needs to start with the name of the function.

And according to the godoc recommendations, the deprecation notice should include a colon:

> To signal that an identifier should not be used, add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation.
>
> https://go.dev/blog/godoc

```suggestion
// DeprecatedRegister adds API health-related endpoints to a Gin-Gonic engine.
//
// Deprecated: Not part of the /api group for endpoints. Tentatively planned
// for complete removal in v6.
```

jilleJr (changes requested):
Some linting suggestions, and I still think that even the new endpoints should allow anonymous access

jilleJr (approved):
LGTM! Remember to merge in changes from master. As you've moved a code chunk in the main.go/main function, you need to be careful when resolving the merge conflicts.

main.go
```diff
@@ -96,30 +116,11 @@ func main() {
 		r.Use(cors.New(corsConfig))
 	}
 
-	healthModule{}.Register(r)
+  healthModule{}.DeprecatedRegister(r)
+  healthModule{}.Register(r.Group("/api"))
```
jilleJr, line 120: nit: Invalid indentation

```suggestion
	healthModule{}.DeprecatedRegister(r)
	healthModule{}.Register(r.Group("/api"))
```

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/iver-wharf/wharf-api
 go 1.16
 
 require (
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
```
jilleJr, line 6: Find it funny how this dependency comes and goes from PR to PR. I have yet to figure out why it is added sometimes and why it sometimes is removed when I run `go mod tidy`, but it seems consistent no matter how many times I run `go mod tidy` in sequence, but a day later and it works the opposite way.

So weird.

jilleJr (approved):
LGTM!

jilleJr (approved):

go.mod
Alexamakans: I have found myself thinking the same thing on numerous occasions...

fredx30 (approved):

PR #46, opened 2021-08-20
Documented remaining values

jilleJr:

## Summary

- Changed a lot of types & functions from exported to unexported.

- Documented the remaining exported types that needs to be exported, such as JSON responses and DB models.

## Motivation

We start growing error fatigue due to the existing errors. Broken window principle applies here.

If we have properly commented everything, it'll be easier to see how we should keep it, and it's be less insentive to just ignore the linting errors.

### DIFF
--- a/database_models.go
+++ b/database_models.go
@@ -19,6 +19,9 @@ const (
 	providerFieldTokenID   = "TokenID"
 )
 
+// Provider holds metadata about a connection to a remote provider. Some of
+// importance are the URL field of where to find the remote, and the token field
+// used to authenticate.
 type Provider struct {
 	ProviderID uint   `gorm:"primaryKey" json:"providerId"`
 	Name       string `gorm:"size:20;not null" json:"name" enum:"azuredevops,gitlab,github"`
@@ -33,6 +36,7 @@ const (
 	tokenFieldUserName = "UserName"
 )
 
+// Token holds credentials for a remote provider.
 type Token struct {
 	TokenID  uint   `gorm:"primaryKey" json:"tokenId"`
 	Token    string `gorm:"size:500; not null" json:"token" format:"password"`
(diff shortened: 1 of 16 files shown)


Alexamakans:
Nice!

I noticed that even if the type is unexported swaggo still generates a file for it, so e.g. `TestsResults` could be `testsResults`.
I seemed to have no issues when doing that, at least.

jilleJr:
> Nice!
> 
> I noticed that swaggo still generates a file for unexported types as well, so e.g. `TestsResults` could be `testsResults`.
> I seemed to have no issues when doing that, at least.

Well I'll file that under "better to have them documented than not", as I find it quite nice to have the comments in the swagger spec, ex:

Screenshot from 2021-08-20 15-30-07

Alexamakans:
> > Nice!
> > I noticed that swaggo still generates a file for unexported types as well, so e.g. `TestsResults` could be `testsResults`.
> > I seemed to have no issues when doing that, at least.
> 
> Well I'll file that under "better to have them documented than not", as I find it quite nice to have the comments in the swagger spec, ex:
> 
> Screenshot from 2021-08-20 15-30-07

Ah! Didn't know it showed up there. Nice.

database_models.go
```diff
@@ -129,6 +142,7 @@ type Log struct {
 	Timestamp time.Time `gorm:"not null" json:"timestamp" format:"date-time"`
 }
 
+// Param holds the definition of a input parameter for a project.
```
Alexamakans, line 145: typo
```suggestion
// Param holds the definition of an input parameter for a project.
```

project.go
```diff
@@ -23,22 +23,25 @@ import (
...
 )
 
-type ProjectModule struct {
+type projectModule struct {
 	Database     *gorm.DB
 	MessageQueue *messagebus.MQConnection
 	Config       *Config
 }
 
+// PaginatedBuilds is a list of builds as well as an explicit total cound field.
```
Alexamakans, line 32: typo
```suggestion
// PaginatedBuilds is a list of builds as well as an explicit total count field.
```

Alexamakans (approved):
2 typos, rest LGTM!

iverestefans (approved):
LGTM!

PR #47, opened 2021-08-24
Moved httputils & updated its logging

jilleJr:

## Summary

Moved httputils to have it placed in the same location as in the provider repos. That is, in the internals dir

## Motivation

To make things consistent

### DIFF
--- a/internal/httputils/httpclient.go
+++ b/internal/httputils/httpclient.go
@@ -10,7 +10,7 @@ import (
 	"github.com/iver-wharf/wharf-core/pkg/logger"
 )
 
-var log = logger.New()
+var log = logger.NewScoped("HTTPUTILS")
 
 // NewClientWithCerts creates a fresh net/http.Client populated with some
 // root CA certificates from file.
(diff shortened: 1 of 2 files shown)


Alexamakans (approved):

iverestefans (approved):

internal/httputils/httpclient.go
```diff
@@ -34,12 +34,11 @@ func NewClientWithCerts(localCertFile string) (*http.Client, error) {
 		return nil, fmt.Errorf("failed to append %q to RootCAs: %v", localCertFile, err)
 	}
 
-	log.Info().WithString("certsFile", localCertFile).Message("Loaded certs from file.")
+	log.Debug().WithString("file", localCertFile).Message("Loaded certs.")
```
fredx30, line 37: Whats the `certsFile` -> `file` change?

fredx30 (approved):

internal/httputils/httpclient.go
jilleJr: I accidentally wrote this logging twice for the httpclient.go. Once here in the wharf-api a while ago, and once again over at https://github.com/iver-wharf/wharf-provider-github/pull/28/files#diff-84aff86cac1871b457a422821d91396633efa081a9d3dd9e3d797dd63ff470aa

The wharf-provider-github's httpclient.go was copied over to the other providers, and only after publishing the PRs I noticed that I also had this in the wharf-api repo but where it already had wharf-core logging.

This change is to make the logging identical across the repos.

*(yea yea I know, this code should be moved to wharf-core, but I didn't find the energy. I will open a new PR in wharf-core ASAP for it though, so we can remove this duplication)*

PR #54, opened 2021-08-31
Fixed project & build JSON marshal regression

jilleJr:

## Summary

- Fixed regression introduced in #46 where `Project.MarshalJSON` and `Build.MarshalJSON` were renamed, unknowing that it broke the Go interface [`Marshaler`](https://pkg.go.dev/encoding/json#Marshaler) from the `encoding/json` package.

  This lead to the `GET /projects` and `GET /project/{projectid}` to not have their parsed build definitions, that are needed in the wharf-web.

- Fixed bug where if it silently ignored failing status parsing

## Motivation

This still regards unreleased changes, so I've not added a dedicated release note for it.

The refactoring was made to try and make it clearer that this marshalling is being done, as @Alexamakans spent hours trying to figure out why the `build` property was missing from the HTTP responses. Hopefully it's made clearer now and could save future devs some hours of debugging.

The best would be to have different structs for database and HTTP responses, but I'm delaying that for the iver-wharf/rfcs#16 implementation as it plans on tackling this.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -14,22 +14,6 @@ func contains(items []string, value string) bool {
 	return false
 }
 
-func unmarshalledYAMLToMarshallableJSON(i interface{}) interface{} {
-	switch x := i.(type) {
-	case map[interface{}]interface{}:
-		m2 := map[string]interface{}{}
-		for k, v := range x {
-			m2[k.(string)] = unmarshalledYAMLToMarshallableJSON(v)
-		}
-		return m2
-	case []interface{}:
-		for i, v := range x {
-			x[i] = unmarshalledYAMLToMarshallableJSON(v)
-		}
-	}
-	return i
-}
-
 func getProjectGroupFromGitURL(gitURL string, projectName string) string {
 	pattern := regexp.MustCompile(`((git@[\w\.]+)):(?P<projectPath>[\w\.@\:/\-~]+)(\.git)(/)?`)
 	if !pattern.MatchString(gitURL) {
--- a/utils.go
+++ b/utils.go
@@ -14,17 +14,17 @@ func contains(items []string, value string) bool {
 	return false
 }
 
-func convert(i interface{}) interface{} {
+func unmarshalledYAMLToMarshallableJSON(i interface{}) interface{} {
 	switch x := i.(type) {
 	case map[interface{}]interface{}:
 		m2 := map[string]interface{}{}
 		for k, v := range x {
-			m2[k.(string)] = convert(v)
+			m2[k.(string)] = unmarshalledYAMLToMarshallableJSON(v)
 		}
 		return m2
 	case []interface{}:
 		for i, v := range x {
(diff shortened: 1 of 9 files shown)


Alexamakans (approved):

utils.go
```diff
@@ -14,17 +14,17 @@ func contains(items []string, value string) bool {
...
+			m2[k.(string)] = unmarshalledYAMLToMarshallableJSON(v)
 		}
 		return m2
 	case []interface{}:
 		for i, v := range x {
-			x[i] = convert(v)
+			x[i] = unmarshalledYAMLToMarshallableJSON(v)
 		}
 	}
 	return i
```
fredx30, line 30: Cool but kinda makes my head hurt. Can you elaborate on why this is needed? There is a yaml to json function and an unmarshal in `github.com/ghodss/yaml`.

jilleJr: Awesome finding! I'll try to use that one instead

jilleJr: The `github.com/ghodss/yaml` doesn't support non-string map keys, so this function is useless. The type will never be `map[interface{}]interface{}`.

The YAML parser used in wharf-core, `gopkg.in/yaml.v2`, does support non-string map keys.

Example:

```go
package main

import (
	"fmt"

	wharfAPIYAML "github.com/ghodss/yaml"
	wharfCoreYAML "gopkg.in/yaml.v2"
)

func main() {
	var yamlBytes = []byte(`123: foo`)

	var obj1 interface{}
	wharfAPIYAML.Unmarshal(yamlBytes, &obj1)
	fmt.Printf("%#v\n", obj1)

	var obj2 interface{}
	wharfCoreYAML.Unmarshal(yamlBytes, &obj2)
	fmt.Printf("%#v\n", obj2)
}

// Output:
// map[string]interface {}{"123":"foo"}
// map[interface {}]interface {}{123:"foo"}
```

This function is older than this repo, so my only guess is that there was perhaps a dependency switch, but the function was not removed.

---

Q: Do we either:

1. Remove `unmarshalledYAMLToMarshallableJSON` as `github.com/ghodss/yaml` doesn't need it?

   The use case here is just to parse the build configuration so the frontend can use a pre-parsed JSON object instead of having to do any YAML parsing itself. As JSON doesn't support non-string keys, this "drawback" from `github.com/ghodss/yaml` could actually be seen as a strength.

2. Refactor `unmarshalledYAMLToMarshallableJSON` to be easier to understand and switch to `gopkg.in/yaml.v2`?

   Better to use the same YAML serializer throughout, instead of having multiple in the same project.

Alexamakans: I would prefer 1.

jilleJr: I've gone for option 1 and added some tests to verify that it works. I also changed the MarshalJSON to be a non-pointer method as that did otherwise only work for pointer values.

fredx30 (approved):
Good stuffs!

PR #64, opened 2021-09-13
Fixed cannot delete projects

### LINKED ISSUE #62: Cannot delete projects
## Expected

```sh
curl -X DELETE 'https://your-wharf-instance.com/api/project/5'
>>> OK
```

And all branches and builds, and their associate build logs and build params, are deleted as well.

As it's quite the dangerous endpoint, maybe we want to add something like `?are-you-sure=true`, or is that not necessary as it's an API anyway, meant to be used by computers?

## Actual

Response given:

```json
{
  "type": "https://iver-wharf.github.io/#/prob/api/unexpected-db-read-error",
  "title": "Error writing to database.",
  "status": 502,
  "detail": "Failed deleting project with ID 5 from database.",
  "instance": "/api/project/5",
  "errors": [
    "ERROR: update or delete on table \"project\" violates foreign key constraint \"fk_branch_project\" on table \"branch\" (SQLSTATE 23503)"
  ]
}
```

You must delete all dependent rows from the other tables `branch` and `build` (and in turn the `log` and `build_param` tables as well) before you can delete the project.

As there's no endpoints to delete them you must do it via SQL commands.

Alexamakans:

## Summary

Changed the constraint behavior between Project and all its children (Branch, Build, BuildParam, Log, Artifact) to `CASCADE` instead of `RESTRICT`, so that we can delete/update a Project without having to do it to the data that is lower in the hierarchy manually.

## Motivation

Closes #62

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 - Fixed where wharf-core logging for Gin debug and error messages were set up
   after they were initially used, leading to a mix of wharf-core and Gin
-  formatted logs. (#63
+  formatted logs. (#63)
 
 ## v4.2.0 (2021-09-10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 ## v4.2.1 (WIP)
 
-- Fixed bug not being able to delete projects without first deleting all child
+- Fixed bug where unable to delete a Project without first deleting all child
   objects. (#64)
 
 ## v4.2.0 (2021-09-10)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 ## v4.2.1 (WIP)
 
-- Fixed not being able to delete projects without first deleting all child
-  objects. (#?)
+- Fixed bug not being able to delete projects without first deleting all child
+  objects. (#64)
 
 ## v4.2.0 (2021-09-10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 	https://changelog.md/
 -->
 
+## v4.2.1 (WIP)
+
+- Fixed not being able to delete projects without first deleting all child
+  objects. (#?)
+
 ## v4.2.0 (2021-09-10)
 
 - Added support for the TZ environment variable (setting timezones ex.
--- a/database_models.go
+++ b/database_models.go
@@ -85,7 +85,7 @@ const (
 type Branch struct {
 	BranchID  uint     `gorm:"primaryKey" json:"branchId"`
 	ProjectID uint     `gorm:"not null;index:branch_idx_project_id" json:"projectId"`
-	Project   *Project `gorm:"foreignKey:ProjectID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+	Project   *Project `gorm:"foreignKey:ProjectID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE" json:"-"`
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,9 +14,12 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 ## v4.2.1 (WIP)
 
+- Fixed bug where unable to delete a Project without first deleting all child
+  objects. (#64)
+
 - Fixed where wharf-core logging for Gin debug and error messages were set up
   after they were initially used, leading to a mix of wharf-core and Gin
-  formatted logs. (#63)
+  formatted logs. (#63
```
fredx30, line 22: nit: 
```suggestion
  formatted logs. (#63)
```

fredx30 (approved):

jilleJr (approved):

migrations.go
```diff
@@ -12,6 +12,21 @@ func runDatabaseMigrations(db *gorm.DB) error {
...
+		{"artifact", "fk_artifact_build"},
+		{"log", "fk_log_build"},
+		{"build", "fk_build_project"},
+		{"log", "fk_log_build"},
+		{"branch", "fk_project_branches"},
+		{"build_param", "fk_build_params"},
+	}); err != nil {
+		return err
+	}
+
```
jilleJr, line 29: So for each migration, which is run every time the wharf-api restarts, the above constraints are removed for a short period.

I assume you added this because GORM doesn't alter the existing constraints, yes?

Highly suggest to instead to take a more surgical approach. Maybe time to start using yet another table for manual migration steps. A table with either a list of migrations applied, or just a single row with an integer that specifies a "migration revision" that the migrations look for. Thoughts?

jilleJr (changes requested):

migrations.go
jilleJr: Or if you could do it in a single transaction for deleting those constraints and then doing the "AutoMigrate" when db.DisableForeignKeyConstraintWhenMigrating is disabled

CHANGELOG.md
Alexamakans: Oops!

migrations.go
Alexamakans: Transitioning to a migrations table would be a good idea for another PR I think. That would be a lot more flexible and maintainable.

I will change it to do it in a single transaction 👍

database_models.go
```diff
@@ -132,7 +132,7 @@ type Build struct {
 type BuildParam struct {
 	BuildParamID uint   `gorm:"primaryKey" json:"-"`
 	BuildID      uint   `gorm:"not null;index:buildparam_idx_build_id" json:"buildId"`
-	Build        *Build `gorm:"foreignKey:BuildID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+	Build        *Build `gorm:"foreignKey:BuildID" json:"-"`
```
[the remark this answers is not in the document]
Alexamakans: I can test with specifying for both and if it still works then sure.

migrations.go
Alexamakans: Done in 8a3a54d

database_models.go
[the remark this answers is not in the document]
Alexamakans: Done in bb1a94f

migrations.go
```diff
@@ -17,9 +17,28 @@ func runDatabaseMigrations(db *gorm.DB) error {
...
+		}
+
+		tx.DisableForeignKeyConstraintWhenMigrating = false
+		if err := tx.AutoMigrate(tables...); err != nil {
+			return fmt.Errorf("migrating with constraints: %w", err)
+		}
+
+		return nil
+	}); err != nil {
+		return fmt.Errorf("transaction failed: %w", err)
```
jilleJr, line 41: nit: To not produce double-phrasing in the error message, I think it's safe to just return the error as-is here.

In my mind, at least, it's the job of `db.Transaction` to say that `db.Transaction` failed. If we are going to extend the error message then we should supply it with context that itself would not be able to provide.

As the code paths inside the transaction func all return pretty detailed errors then I vote for just doing this:

```suggestion
		return err
```

jilleJr (approved):

PR #77, opened 2021-09-16
Fixed some things missed in PR #43

Alexamakans:

## Summary

Changed the comment for `artifactModule.getBuildTestsResultsHandler` so that it refers to the correct endpoint.

Fixed a typo in `TestResultDetail` field tags, `enums` changed to `enum`.

Fixed constraint behavior definition for `Build.TestResultSummaries`.

## Motivation

Being pointed to a non-existing endpoint when something is deprecated isn't a great thing, this prevents anybody experiencing that.

The typo made it so the `status` field generated over at [iver-wharf/wharf-api](https://github.com/iver-wharf/wharf-api) was of type `MainTestResultDetail` instead of `string`, causing some issues with simple integration.

The constraint behavior definition was only defined on `TestResultSummary.Build`, which meant that it was set to the default `RESTRICT`, not `CASCADE` that it should be.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
   Use `GET /build/{buildid}/test-result/list-summary` instead. The response
   data is slightly different; it has additional properties, and does not have a
-  `status` property. (#43)
+  `status` property. (#43, #77)
 
 ## v4.2.0 (2021-09-10)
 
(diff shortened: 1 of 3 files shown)


jilleJr (approved):

jilleJr (approved):

Alexamakans:
Added `bug` label because the constraint behavior definition was not applied properly. This made it impossible to delete builds that had `TestResultSummary`s connected to them, and because of that also projects with that kind of build.

fredx30 (approved):

Issue #72, opened 2021-09-15
Split up DB & HTTP requests+response models into packages


jilleJr:
> Based on [RFC-0016](https://iver-wharf.github.io/rfcs/published/0016-wharf-api-endpoints-cleanup) (https://github.com/iver-wharf/rfcs/pull/16)

Explained in the RFC:

> Starting with v5, the wharf-api no longer accepts nor returns database models from ANY of its endpoints. Instead, there are now 3 new packages of models inside the wharf-api:
>
> - `pkg/models/database`: /.../
> - `pkg/models/request`: /.../
> - `pkg/models/response`: /.../
>
> https://iver-wharf.github.io/rfcs/published/0016-wharf-api-endpoints-cleanup#requestresponse-models-changes

jilleJr:
Decided to go with `pkg/model/...` instead of `pkg/models/...` to stick with the singular package naming convention

Issue #71, opened 2021-09-15
Replace "POST search"+JSON body with "GET"+query params


jilleJr:
> Based on [RFC-0016](https://iver-wharf.github.io/rfcs/published/0016-wharf-api-endpoints-cleanup) (https://github.com/iver-wharf/rfcs/pull/16)

As discussed in the "Alternative solutions":

> While [using POST for search endpoints] can be useful, it’s not required for today’s use cases. Not banning POST searches for future use, but for these simpler search queries they do not fit well.
>
> https://iver-wharf.github.io/rfcs/published/0016-wharf-api-endpoints-cleanup#alternative-solutions

## Expected

```http
GET /project/{projectId}/build
```

```
 NAME              PARAM    TYPE               REQUIRED?  DESCRIPTION
 projectId         (path)   integer            true
 limit             (query)  integer            false      Max number of items returned.
 offset            (query)  integer            false      Shifts the window returned.
 orderby           (query)  array[string]      false      Alphabetically, or order by ID?
 environment       (query)  string             false      Filter on environment hard match
 environmentMatch  (query)  string             false      Filter on environment soft match
 finishedAfter     (query)  string[date-time]  false      Filter on finishedOn
 finishedBefore    (query)  string[date-time]  false      Filter on finishedOn
 gitBranch         (query)  string             false      Filter on gitBranch hard match
 gitBranchMatch    (query)  string             false      Filter on gitBranch soft match
 isInvalid         (query)  boolean            false      Filter on isInvalid
 scheduledAfter    (query)  string[date-time]  false      Filter on scheduledOn
 scheduledBefore   (query)  string[date-time]  false      Filter on scheduledOn
 stage             (query)  string             false      Filter on stage hard match
 stageMatch        (query)  string             false      Filter on stage soft match
 status            (query)  string[enum]       false      Filter on status by enum string
 statusId          (query)  integer            false      Filter on status by ID
```

The difference between a soft match (params ending with `-Match`) vs hard match is that:

- hard match: verbatim match on the whole string. Useful when other components or cronjobs want to search for items.
- soft match: query can be a substring of the real value, or even allow typos. Useful when humans want to search for items.

Idea is that if the query parameter is not used, then it will not be searched on. While if it's used then even if the parameter is empty then it's searched on (only relevant for the hard-match filters). Which means that this:

```http
GET /project/{projectId}/build?stage=
```

...resolves to:

```sql
[...]

PR #78, opened 2021-09-16
Split up DB models: part 1

jilleJr:

## Summary

- Added packages
  - `pkg/model/database`
  - `pkg/model/request`
  - `pkg/model/response`
- Changed usage of `{type}Field{FieldName}` to `database.{Type}Fields.{FieldName}`
- Updated `token.go` to use new models
- Updated migrations to use new models
- Added transaction to PUT token and made it associate with the provider by the ID from the request JSON.

## Motivation

Related to #72, but doesn't close it yet as there are a lot more files to be dealt with.

This is like a taste sample. Better to get feedback on this early, as changing the remaining files will result in a giant PR that does pretty much the same stuff everywhere.

The -Search model will remain until we drop the backward compatibility, but only be used in the endpoints that will be deprecated in #71.

## Preview

The Swagger documentation now specifies more relevant models. I played around with the [`validate` tag from Swaggo](https://github.com/swaggo/swag#available) as well to mark some fields as required (the small red asterisk):

Screenshot from 2021-09-16 17-34-30
Screenshot from 2021-09-16 17-35-03
Screenshot from 2021-09-16 17-54-05

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 - Added packages for "Plain Old Go Objects", with finer-grained decoupling
   between database, HTTP request, and HTTP response models.
-  The Swagger documentation is affected by this, and some unused fields has
+  The Swagger documentation is affected by this, and some unused fields have
   been removed from certain endpoints, such as the `tokenId` in `POST /token`.
   The new packages are: (#78)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,20 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
   data is slightly different; it has additional properties, and does not have a
   `status` property. (#43)
 
+- Added packages for "Plain Old Go Objects", with finer-grained decoupling
+  between database, HTTP request, and HTTP response models.
+  The Swagger documentation is affected by this, and some unused fields has
+  been removed from certain endpoints, such as the `tokenId` in `POST /token`.
+  The new packages are: (#78)
+
+  - `pkg/model/database`
+  - `pkg/model/request`
+  - `pkg/model/response`
+
+- Fixed `PUT /token` where it did not use the `providerId` value from the HTTP
+  request body. It now sets the provider's token if the field is supplied and
+  non-zero. (#78)
+
 ## v4.2.0 (2021-09-10)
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,20 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
   data is slightly different; it has additional properties, and does not have a
   `status` property. (#43)
 
+- Added packages for "Plain Old Go Objects", with finer-grained decoupling
+  between database, HTTP request, and HTTP response models.
+  The Swagger documentation is affected by this, and some unused fields has
```
Alexamakans, line 54:
```suggestion
  The Swagger documentation is affected by this, and some unused fields have
```

Alexamakans (approved):
Good stuff 👍🏻 
Extra nice with the validations too 👍🏻

iverestefans (commented):
LGTM!

fredx30 (commented):
I like this idea of adding the required asterixes to swaggo. 
The idea of splitting up db models should terrify just about anyone. I dont see any explanation as to why this is a good idea. Elaborate?

jilleJr:
> @fredx30  I like this idea of adding the required asterixes to swaggo.
> The idea of splitting up db models should terrify just about anyone. I dont see any explanation as to why this is a good idea. Elaborate?

Mentioned in the RFC's motivation. One of the points are this, which is relevant:

> - We use the database models in our HTTP request and response specs. The problems with this are:
>
>    - Security weakness as it induces a higher risk of exposing too much data to the user.
>
>   - Swagger documentation has room for improvement. We can elide irrelevant specifications such as allowing a project ID to be provided when creating a new project.
>
>   - The abstraction layers get mixed. With different models we would decouple the HTTP layer from the database layer, allowing for easier customization in both layers, as well as easier to understand as the models doesn’t have to worry both about how to serialize into JSON as well as their database table relations and constraints.
>
> https://iver-wharf.github.io/rfcs/published/0016-wharf-api-endpoints-cleanup#motivation

jilleJr:
@iverestefans you only left a comment and not an "approve" review :)

jilleJr:
@Alexamakans re-requested your review because of 461f49e & 45d00dd which was pushed after your review

Alexamakans (approved):

iverestefans (approved):
LGTM!

PR #80, opened 2021-09-21
Fixed N+1 Build and TestResult fetch issue

Alexamakans:

## Summary

Added `Build.TestResultListSummary` field.
This field is manually populated when fetching a `Build`, using the function `populateTestResultListSummary` located in `build.go`.

Changed so `Build.TestResultSummaries` is preloaded in `buildModule.getBuild` and `projectModule.getBuilds`.

## Motivation

One big part of the test result handling rework (#43) was to fix N+1 Select issue(s) in [wharf-web](https://github.com/iver-wharf/wharf-web), but I seem to have mucked up the ability to do that. This should fix that.

## Note

I wasn't able to think of a way to populate `Build.TestResultListSummary` in the same DB request as getting them (`buildModule.getBuild` and `projectModule.getBuilds`). We could add `TestResultListSummary` to the database, but since it's just a calculated value I'm not sure about if that's the way we should do it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,12 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
   ID has not been shown to have any benefits. Please refer to the
   `GET /project/{projectId}` endpoint instead. (#75)
 
+- Added `TestResultListSummary` field to `Build` database model. This should
+  allow us to avoid `N+1 Select` in wharf-api. (#80)
+
+- Changed to preload `TestResultSummaries` field of `Build` database
+  model. (#80)
+
 ## v4.2.0 (2021-09-10)
 
 - Added support for the TZ environment variable (setting timezones ex.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,8 +64,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
   ID has not been shown to have any benefits. Please refer to the
   `GET /project/{projectId}` endpoint instead. (#75)
 
-- Added `TestResultListSummary` field to `Build` database model. This should
-  allow us to avoid `N+1 Select` in wharf-api. (#80)
+- Added `TestResultListSummary` field to `Build` database model. This allows you
+  to avoid `N+1` HTTP requests when listing builds to show test summaries. (#80)
 
 - Changed to preload `TestResultSummaries` field of `Build` database
   model. (#80)
--- a/build.go
+++ b/build.go
@@ -132,7 +132,7 @@ func (m buildModule) getBuild(buildID uint) (Build, error) {
 	var build Build
 	if err := m.Database.
 		Where(&Build{BuildID: buildID}).
-    Preload(database.BuildFields.TestResultSummaries).
+		Preload(database.BuildFields.TestResultSummaries).
 		Preload(database.BuildFields.Params).
 		First(&build).
 		Error; err != nil {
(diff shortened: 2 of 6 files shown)


build.go
```diff
@@ -360,3 +364,20 @@ func setStatusDate(build *Build, statusID BuildStatus) {
...
+		Scan(listSummaryPtr).
+		Error; err != nil {
+		return err
+	}
+	listSummaryPtr.Total =
+		listSummaryPtr.Failed +
+			listSummaryPtr.Passed +
+			listSummaryPtr.Skipped
+	return nil
+}
```
jilleJr, line 383: Would rather see this function use return value instead of mutating the argument.

```go
summary, err := getTestResultListSummary(m.Database, buildID)
if err != nil {
	return Build{}, err
}
build.TestResultListSummary = summary

// ...

func getTestResultListSummary(db *gorm.DB, buildID uint, TestResultListSummary) error { }
```

It's a personal preference to prefer "pure functions" (non-mutating functions) as I've had better experience with it and felt that it produces less bugs. Mutating the argument is easily unseen when reading the code, except in some cases such as in `Marshal` and `Unmarshal` functions

CHANGELOG.md
```diff
@@ -64,6 +64,12 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
   ID has not been shown to have any benefits. Please refer to the
   `GET /project/{projectId}` endpoint instead. (#75)
 
+- Added `TestResultListSummary` field to `Build` database model. This should
+  allow us to avoid `N+1 Select` in wharf-api. (#80)
```
jilleJr, line 68: The implementation in `getBuilds` inside `project.go` still is a `N+1` implementation for the SQL SELECTs.

I'd say it's OK for now, especially as users typically query at most 10 builds at a time, but I'd suggest changing this phrasing

```suggestion
- Added `TestResultListSummary` field to `Build` database model. This allow you
  to avoid `N+1` HTTP requests when listing builds to show test summaries. (#80)
```

jilleJr (changes requested):

jilleJr:
About your note in the description:

> I wasn't able to think of a way to populate `Build.TestResultListSummary` in the same DB request as getting them (`buildModule.getBuild` and `projectModule.getBuilds`). We could add `TestResultListSummary` to the database, but since it's just a calculated value I'm not sure about if that's the way we should do it.

I'd say it's fine for now, as mentioned in https://github.com/iver-wharf/wharf-api/pull/80#discussion_r716435214, as this at least allows the client to not have to do any N+1 requests. The server still makes N+1 requests, but since the builds are paginated then I conclude that it's OK until we see performance issues, or when we have users who want to query 1000 builds at a time or something.

build.go
```diff
@@ -132,11 +132,17 @@ func (m buildModule) getBuild(buildID uint) (Build, error) {
 	var build Build
 	if err := m.Database.
 		Where(&Build{BuildID: buildID}).
+    Preload(database.BuildFields.TestResultSummaries).
```
fredx30, line 135: Indent appears to be off.

jilleJr, line 135: nit: indentation bo-bo

```suggestion
		Preload(database.BuildFields.TestResultSummaries).
```

jilleJr (approved):

fredx30 (approved):
The note in the description seems indicative of the need for a discussion regarding the choice of solution. I think you should have booked that as a call before creating the PR. Such a discussion would make it easier to review this.

build.go
jilleJr: ha! beat you to it!

PR #76, opened 2021-09-15
Changed path params to camelCase

jilleJr:
- \[x] I've added a new note in the `CHANGELOG.md` file, according to docs:
  https://iver-wharf.github.io/#/development/changelogs/writing-changelogs

## Summary

- `branchid` &rarr; `branchId`
- `projectid` &rarr; `projectId`
- `providerid` &rarr; `providerId`
- `tokenid` &rarr; `tokenId`
- `buildid` &rarr; `buildId`

I did a search and replace on all files inside the repo. No all-lowercase alternatives, except for inside the CHANGELOG.md, remain.

## Motivation

Closes #68

PR #82, opened 2021-09-22
Removed Provider.UploadURL

### LINKED ISSUE #56: Remove Provider.UploadURL
Old and unused field on the provider model.

For GitHub, the API URL is https://api.github.com/, but to upload assets to a release you use https://uploads.github.com. This URL is however provided by the ["Create a release"](https://docs.github.com/en/rest/reference/repos#upload-a-release-asset) endpoint and should be able to be provided from there instead.

I've not researched the other providers, but I do know that the upload URL field is unused throught Wharf.

Migrations are needed. GORM doesn't removes columns automatically, it only adds them automatically.

Alexamakans:

## Summary

Removed field `Provider.UploadURL` and all references to it.
Added migration to remove DB column `provider.upload_url`

## Motivation

This field was unused and mostly lead to confusion.

## Notes

Tried building wharf-web after the change but it wouldn't pass because `MainProvider.uploadUrl` from the previous swag definitions was used at `src/app/providers/providers.service.ts:L50`.

The change of `id` -> `Id` from #76 also causes the build to fail because of the auto-generated names for the endpoint functions, so I believe a MAJOR version bump is in order.

--
Closes #56.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 	https://changelog.md/
 -->
 
-## v4.3.0 (WIP)
+## v5.0.0 (WIP)
 
 - Fixed bug where unable to delete a Project without first deleting all child
   objects. (#64)
@@ -63,6 +63,10 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 - Deprecated endpoint `GET /branch/{branchId}`. Getting a single branch by its
   ID has not been shown to have any benefits. Please refer to the
   `GET /project/{projectId}` endpoint instead. (#75)
+
+- Removed `Provider.UploadURL` and all references to it, as it was unused. (#82)
+
+- Removed DB column `provider.upload_url`, as it was unused. (#82)
 
 ## v4.2.0 (2021-09-10)
 
--- a/database_models.go
+++ b/database_models.go
@@ -26,10 +26,9 @@ import (
 // consistent we add it to both referencing fields.
 
 const (
-	providerFieldName      = "Name"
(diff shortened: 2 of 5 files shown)


Alexamakans:
Both [security alerts](https://github.com/iver-wharf/wharf-api/pull/82/checks?check_run_id=3673194667) are valid I think, related to https://github.com/iver-wharf/wharf-provider-azuredevops/issues/19, but for database queries.

They are validated to be fully qualified URLs in the provider repos, but since this validation happens outside of the wharf-api repo there is nothing stopping anybody from crafting their own request.

jilleJr:
> Both [security alerts](https://github.com/iver-wharf/wharf-api/pull/82/checks?check_run_id=3673194667) are valid I think, related to [iver-wharf/wharf-provider-azuredevops#19](https://github.com/iver-wharf/wharf-provider-azuredevops/issues/19), but for database queries.
> 
> They are validated to be fully qualified URLs in the provider repos, but since this validation happens outside of the wharf-api repo there is nothing stopping anybody from crafting their own request.

The security alerts are about SQL injections. GORM does not have any known SQL injection issues, as it escapes all values for us. From the "Recommendation" section of the alert:

> /.../ Use these features rather than building queries by string concatenation.
>
> https://github.com/iver-wharf/wharf-api/security/code-scanning/33?query=ref%3Arefs%2Fpull%2F82%2Fmerge+ref%3Arefs%2Fpull%2F82%2Fhead+ref%3Arefs%2Fheads%2Ffeature%2Fremove-provider-upload-url

We're not concatenating any strings... Those with Rule ID `go/sql-injection` are false-positives

jilleJr (approved):

fredx30 (approved):

PR #100, opened 2021-10-22
Migrate columns to `NOT NULL`

jilleJr:

## Summary

- Added custom migration code for migrating columns in PostgreSQL to add `NOT NULL`
- Changed all database fields that were non-pointers from `gorm:"nullable"` to `gorm:"not null;default:''"`

## Motivation

This adds consistency between the wharf-api and the database.

Some fields cannot be null, according to our database models, as we're using `string` and not `*string`. Because of this, GORM will never set the values to `NULL` in the database as it's solely relying on the value in Go. In other words: for all the changed fields in the database models, none *should* actually have the value `NULL` in the database, but instead only empty strings.

So the migration steps might be overkill, but I'd rather have them than not, just to catch those edge cases where someone might have updated a field via raw SQL manually, or if some previous migration added nullable columns.

The Postgres GORM migrator implementation could only change a column to nullable, but not from nullable to `NOT NULL`, probably because that's an errornous operation and requires default values. This new implementation does supply those default values.

### DIFF
--- a/migrations.go
+++ b/migrations.go
@@ -29,6 +29,8 @@ func runDatabaseMigrations(db *gorm.DB, driver DBDriver) error {
 				" We advice against using this driver for production!")
 	}
 
+	// since v5.0.0, all columns that were not nil'able in the GORM models
+	// has been migrated to not be nullable in the database either.
 	if err := migrateWharfColumnsToNotNull(driver, db); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


migrations.go
```diff
@@ -28,6 +29,10 @@ func runDatabaseMigrations(db *gorm.DB, driver DBDriver) error {
 				" We advice against using this driver for production!")
 	}
 
+	if err := migrateWharfColumnsToNotNull(driver, db); err != nil {
```
jilleJr, line 32:
```suggestion
	// since v5.0.0, all columns that were not nil'able in the GORM models
	// has been migrated to not be nullable in the database either.
	if err := migrateWharfColumnsToNotNull(driver, db); err != nil {
```

fredx30 (approved):

PR #101, opened 2021-11-01
Allow cred/authorization header CORS

fredx30:

## Summary

This adds a environment variable that allows setting a cors allowed origin. When this is set Authrization Bearer tokens will also be allowed to be sent. This overrides the allow all origins header. When sending credentials allow all origins is not allowed.

## Motivation

When sending credentials allow all origins is not allowed. The way the OICD login is set to work credentials need to be sent and will use the `authorization` header.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,6 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
   - Use `PUT /token/{tokenId}` instead of `PUT /token`
   - Use `PUT /project/{projectId}/branch` instead of `PUT /branches`
 
-
 - Deprecated POST search endpoints that took the search queries from the HTTP
   request body. They are still supported, but may be removed in the next major
   release (v6.0.0). Please refer to the new endpoints that use query parameter
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 ## v5.0.0 (WIP)
 
+- Enables setting specific origins for CORS via the
+  WHARF_HTTP_CORS_ALLOWORIGINS environment variable. This is to make sending
+  `Authorization` headers possible. (#101)
+
 - BREAKING: Removed all deprecated environment variable configs, which were
   marked as deprecated in v4.2.0/#38. Now all environment variables require the
   `WHARF_` prefix. (#87)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,6 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 ## v5.0.0 (WIP)
 
-- Enables setting specific origins for CORS via the
-  WHARF_HTTP_CORS_ALLOWORIGINS environment variable. This is to make sending
-  `Authorization` headers possible. (#101)
-
 - BREAKING: Removed all deprecated environment variable configs, which were
   marked as deprecated in v4.2.0/#38. Now all environment variables require the
   `WHARF_` prefix. (#87)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 ## v5.0.0 (WIP)
 
+- Enables setting specific origins for CORS via the
+  WHARF_HTTP_CORS_ALLOWORIGINS environment variable. This is to make sending
+  `Authorization` headers possible. (#101)
+
```
jilleJr, line 20: Suggest rephrasing this to begin with "Added", perhaps also copy the format used in some other logs, such as:

> ```md
> - Added configuration for Sqlite file path, environment variable `WHARF_DB_PATH`
>   or the YAML key `db.path`. Defaults to `wharf-api.db`. (#86)
> ```

```suggestion
- Added configuration of specific origins for CORS via the environment variable
  `WHARF_HTTP_CORS_ALLOWORIGINS` or the YAML key `http.cors.allowOrigins`. This
  is to make sending `Authorization` headers possible. (#101)
```

config.go
```diff
@@ -122,6 +122,13 @@ type CORSConfig struct {
 	//
 	// Added in v4.2.0.
 	AllowAllOrigins bool
+
+	// AllowOrigins enables CORS and allows the list of origins in the
+	// HTTP request origins when set. Practically speaking, this
+	// results in the HTTP header "AllowOrigins".
```
jilleJr, line 128: Wrong header name :)

```suggestion
	// results in the HTTP header "Access-Control-Allow-Origin".
```

config.go
```diff
@@ -122,6 +122,13 @@ type CORSConfig struct {
...
+	//
+	// Added in v5.0.0.
+	AllowOrigins string
```
jilleJr, line 131: I think this should be a slice of strings. While the header only can include 1 value, IIRC Gin will check the `Origin` header from the request and use the appropriate origin in the response header.

While we probably won't need more than 1, it's good to have the option available, especially since it's such an easy change but hard to change later as it changes the config schema.

Apparently Viper (that we're using for the configs) supports slices by separating elements via a comma.
(https://github.com/spf13/viper/blob/v1.7.1/viper.go#L920)

So you can change this to `[]string` and then add something like this to the godoc:
> Multiple origins can be supplied via environment variables by separating them by commas. YAML configuration is expected to use a regular array of strings.

main.go
```diff
@@ -110,7 +110,14 @@ func main() {
 		ginutil.RecoverProblem,
 	)
 
-	if config.HTTP.CORS.AllowAllOrigins {
+	if len(config.HTTP.CORS.AllowOrigins) > 0 {
+		log.Info().Message("Allowing " + config.HTTP.CORS.AllowOrigins + " origins in CORS.")
```
jilleJr, line 114: nit: A rule of thumb I've used for logs is to not have anything varying in the message.

Suggest moving the config value here to a `WithString` call. For example:

```suggestion
		log.Info().WithString("origin", config.HTTP.CORS.AllowOrigins).
			Message("Allowing origins in CORS.")
```

But if you're changing to a string slice, then perhaps just something like this:

```suggestion
		log.Info().WithStringf("origin", "%v", config.HTTP.CORS.AllowOrigins).
			Message("Allowing origins in CORS.")
```

^ would resolve to something like ``origins=`[http://localhost:5001 http://localhost:4200]` `` in the console output for multiple values

Otherwise a more custom solution:


```suggestion
		log.Info().WithString("origin", strings.Join(config.HTTP.CORS.AllowOrigins, " ")).
			Message("Allowing origins in CORS.")
```

^ would resolve to something like ``origins=`http://localhost:5001 http://localhost:4200` `` in the console output for multiple values

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 
 ## v5.0.0 (WIP)
 
+- Enables setting specific origins for CORS via the
+  WHARF_HTTP_CORS_ALLOWORIGINS environment variable. This is to make sending
+  `Authorization` headers possible. (#101)
```
jilleJr, line 19: Could you place this note below the "BREAKING" notes? I think they should be highest as they're the most important

jilleJr (changes requested):

Alexamakans (changes requested):
I agree that `AllowOrigins` should be `[]string`, with updated godoc.
Prefer opt 2 for logging, resulting in ``origins=`[http://localhost:5001 http://localhost:4200]` ``

Will approve once `AllowOrigins` is changed to `[]string` 👍🏻

Alexamakans (approved):

main.go
fredx30: Sol 2 seems to work

jilleJr (approved):
LGTM!

Issue #69, opened 2021-09-15
Move & rename endpoints


jilleJr:
> Based on [RFC-0016](https://iver-wharf.github.io/rfcs/published/0016-wharf-api-endpoints-cleanup) (https://github.com/iver-wharf/rfcs/pull/16)

For the sake of consistency, all endpoints should follow a more defined format. Currently (in v4.2.0) the endpoints mix between singular and plural.

All endpoints should be in singular.

The full list of proposed changes can be found at https://iver-wharf.github.io/rfcs/published/0016-wharf-api-endpoints-cleanup#renamed-path-parameters

PR #109, opened 2021-11-11
Added remaining GET search endpoints

### LINKED ISSUE #71: Replace "POST search"+JSON body with "GET"+query params

jilleJr:

## Summary

- Added implementation for GET endpoints, based on `GET /project` from #99:

  - `GET /build`
  - `GET /build/{buildId}/artifact`
  - `GET /token`
  - `GET /provider`

- Changed `GET /project` to return `response.PaginatedProjects` instead. All other added GET endpoints also return `response.Paginated...` responses, which includes a `totalCount` in addition to the list of objects, specifying how many items there are that were no matter the Limit & Offset values.

- Deprecated old search and GET endpoints:

  - `GET /projects/{projectId}/builds`, in favor of `GET /build?projectId={projectId}`
  - `POST /builds/search`, in favor of `GET /build`
  - `GET /tokens`, in favor of `GET /token`
  - `POST /tokens/search`, in favor of `GET /token`
  - `GET /providers`, in favor of `GET /provider`
  - `POST /providers/search`, in favor of `GET /provider`

- Added `utils_gorm.go` and `utils_gin.go`, and moved code from `utils.go` to respective files.

- Added `commonGetQueryParams` with Limit, Offset, and OrderBy, to deduplicate code between the new GET endpoints

- Added `findDBPaginatedSliceAndTotalCount` to deduplicate code between the new GET endpoints

## Motivation

Changing from just returning an array of artifacts to the PaginatedArtifacts model is a major breaking change, but luckily all previous GET endpoints were in plural, so their deprecated behavior of only returning arrays is still supported.

Related to #69. I deprecated the endpoints I touched while in the process instead of just upgrading them, but there are still some endpoints left that need to be moved.

Closes #71. All POST search endpoints have now been replaced by GET endpoints.

Sorry for the huge PR. It's a lot of the same as from PR #99, but generalized a bit to deduplicate and applied to all endpoitns.


jilleJr:
Q: For consistency, shall I change the `GET /build/{buildId}/test-result/...` endpoints to return paginated results as well?

Perhaps in a separate PR?

Alexamakans:
> Q: For consistency, shall I change the `GET /build/{buildId}/test-result/...` endpoints to return paginated results as well?
> 
> Perhaps in a separate PR?

Suggest different PR for that, as it is not really the same change as the other endpoints and this one is already big enough IMO. Easy that that change would get lost in the mix I think.

build.go
```diff
@@ -100,40 +102,144 @@ func (m buildModule) getBuildHandler(c *gin.Context) {
...
+// @description Verbatim filters will match on the entire string used to find exact matches,
+// @description while the matching filters are meant for searches by humans where it tries to find soft matches and is therefore inaccurate by nature.
 // @tags build
-// @accept json
-// @produce json
-// @success 501 "Not Implemented"
-// @router /builds/search [post]
-func (m buildModule) searchBuildListHandler(c *gin.Context) {
-	c.Status(http.StatusNotImplemented)
+// @param limit query int false "Number of results to return. No limit if unset or non-positive. Required if `offset` is used." default(100)
```
Alexamakans, line 125: Dupe of previous.

Description here doesn't seem to match the changelog description:
> By default all these new endpoints use a default limit of 100, but this can
  be disabled by specifying `?limit=0`.

Assuming the changelog description is the actual behavior?
```suggestion
// @param limit query int false "Number of results to return. Set to 0 or non-positive to disable. Required if `offset` is used." default(100)
```
Alt.
```suggestion
// @param limit query int false "Number of results to return. Set to 0 to disable. Required if `offset` is used." minimum(0) default(100)
```

project.go
```diff
@@ -65,8 +65,8 @@ var projectJSONToColumns = map[string]string{
 // @description while the matching filters are meant for searches by humans where it tries to find soft matches and is therefore inaccurate by nature.
 // @tags project
 // @param orderby query []string false "Sorting orders. Takes the property name followed by either 'asc' or 'desc'. Can be specified multiple times for more granular sorting. Defaults to `?orderby=projectId desc`"
-// @param limit query int false "Number of results to return. No limit if unset or non-positive. Required if `offset` is used."
-// @param offset query int false "Skipped results, where 0 means from the start." minimum(0)
+// @param limit query int false "Number of results to return. No limit if unset or non-positive. Required if `offset` is used." default(100)
```
Alexamakans, line 68: Dupe of previous.

Description here doesn't seem to match the changelog description:
> By default all these new endpoints use a default limit of 100, but this can
  be disabled by specifying `?limit=0`.

Assuming the changelog description is the actual behavior?
```suggestion
// @param limit query int false "Number of results to return. Set to 0 or non-positive to disable. Required if `offset` is used." default(100)
```
Alt.
```suggestion
// @param limit query int false "Number of results to return. Set to 0 to disable. Required if `offset` is used." minimum(0) default(100)
```

provider.go
```diff
@@ -19,37 +22,100 @@ type providerModule struct {
...
 // @id getProviderList
-// @summary Returns first 100 providers
+// @summary Get slice of providers.
+// @description List all providers, or a window of providers using the `limit` and `offset` query parameters. Allows optional filtering parameters.
+// @description Verbatim filters will match on the entire string used to find exact matches,
+// @description while the matching filters are meant for searches by humans where it tries to find soft matches and is therefore inaccurate by nature.
 // @tags provider
-// @success 200 {object} []response.Provider
-// @failure 401 {object} problem.Response "Unauthorized or missing jwt token"
+// @param limit query int false "Number of results to return. No limit if unset or non-positive. Required if `offset` is used." default(100)
```
Alexamakans, line 53: Dupe of previous.

Description here doesn't seem to match the changelog description:
> By default all these new endpoints use a default limit of 100, but this can
  be disabled by specifying `?limit=0`.

Assuming the changelog description is the actual behavior?
```suggestion
// @param limit query int false "Number of results to return. Set to 0 or non-positive to disable. Required if `offset` is used." default(100)
```
Alt.
```suggestion
// @param limit query int false "Number of results to return. Set to 0 to disable. Required if `offset` is used." minimum(0) default(100)
```

token.go
```diff
@@ -19,38 +22,85 @@ type tokenModule struct {
...
+// getBuildListHandler godoc
 // @id getTokenList
-// @summary Returns first 100 tokens
+// @summary Get slice of tokens.
+// @description List all tokens, or a window of tokens using the `limit` and `offset` query parameters. Allows optional filtering parameters.
+// @description Verbatim filters will match on the entire string used to find exact matches,
+// @description while the matching filters are meant for searches by humans where it tries to find soft matches and is therefore inaccurate by nature.
 // @tags token
-// @success 200 {object} []response.Token
+// @param limit query int false "Number of results to return. No limit if unset or non-positive. Required if `offset` is used." default(100)
```
Alexamakans, line 53: Dupe of previous.

Description here doesn't seem to match the changelog description:
> By default all these new endpoints use a default limit of 100, but this can
  be disabled by specifying `?limit=0`.

Assuming the changelog description is the actual behavior?
```suggestion
// @param limit query int false "Number of results to return. Set to 0 or non-positive to disable. Required if `offset` is used." default(100)
```
Alt.
```suggestion
// @param limit query int false "Number of results to return. Set to 0 to disable. Required if `offset` is used." minimum(0) default(100)
```

pkg/modelconv/artifactconv.go
```diff
@@ -14,3 +14,13 @@ func DBArtifactToResponse(dbArtifact database.Artifact) response.Artifact {
 		FileName:   dbArtifact.FileName,
 	}
 }
+
+// DBArtifactsToResponses converts a slice of database branches to a slice of
+// response branches.
```
Alexamakans, line 19:
```suggestion
// DBArtifactsToResponses converts a slice of database artifacts to a slice of
// response artifacts.
```

pkg/model/response/response.go
```diff
@@ -8,6 +8,19 @@ import (
...
 
+// ArtifactJSONFields holds the JSON field names for each field.
+// Useful in ordering statements to map the correct field to the correct
+// database column.
+var ArtifactJSONFields = struct {
+	ArtifactID string
+	Name       string
+	FileName   string
+}{
+	ArtifactID: "artifactsId",
```
Alexamakans, line 19:
```suggestion
	ArtifactID: "artifactId",
```

utils_gorm.go
```diff
@@ -0,0 +1,181 @@
...
+	"strings"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/iver-wharf/wharf-core/pkg/ginutil"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+func findDBPaginatedSliceAndTotalCount(dbQuery *gorm.DB, limit, offset int, slice interface{}, totalCount *int64) error {
```
Alexamakans, line 15: nit: rename `slice` -> `slicePtr` as that is what is being passed to the function when used throughout the code. Good to be explicit when dealing with `interface{}`.
Not sure if gorm's `Find` deals with it so that either is fine, but feels scary to rely on that.
```suggestion
func findDBPaginatedSliceAndTotalCount(dbQuery *gorm.DB, limit, offset int, slicePtr interface{}, totalCount *int64) error {
```

utils_gorm.go
```diff
@@ -0,0 +1,181 @@
...
+	return true
+}
+
+func optionalLimitOffsetScope(limit, offset int) func(*gorm.DB) *gorm.DB {
+	if limit <= 0 {
+		return gormIdentityScope
+	}
+	if offset <= 0 {
+		offset = 0
+	}
```
Alexamakans, line 50: tiniest nit ever 😄:
```suggestion
	if offset < 0 {
		offset = 0
	}
```

utils_gorm.go
```diff
@@ -0,0 +1,181 @@
...
+	return func(db *gorm.DB) *gorm.DB {
+		return db.Clauses(clause.Limit{
+			Limit:  limit,
+			Offset: offset,
+		})
+	}
+}
+
+func optionalTimeRangeScope(column string, min, max *time.Time) func(*gorm.DB) *gorm.DB {
```
Alexamakans, line 59: Do we want to add a check for `max < min` here, or do we get an error for that when running the query?

Alexamakans (changes requested):
Reviewed most of it, still have `utils_gorm.go` to comb through a bit more.

I think it looks good, and I don't see any major changes that need to be made.
We will notice later whether or not it actually is messy/hard to work in, or if it's just because there's a lot of new stuff going on that it feels a bit overwhelming.

Good job 👍🏻

artifact.go
```diff
@@ -22,47 +24,105 @@ type artifactModule struct {
...
 // getBuildArtifactListHandler godoc
 // @id getBuildArtifactList
 // @summary Get list of build artifacts
+// @description List all build artifacts, or a window of build artifacts using the `limit` and `offset` query parameters. Allows optional filtering parameters.
+// @description Verbatim filters will match on the entire string used to find exact matches,
+// @description while the matching filters are meant for searches by humans where it tries to find soft matches and is therefore inaccurate by nature.
 // @tags artifact
 // @param buildId path uint true "Build ID" minimum(0)
-// @success 200 {object} []response.Artifact
+// @param limit query int false "Number of results to return. No limit if unset or non-positive. Required if `offset` is used." default(100)
```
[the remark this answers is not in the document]
jilleJr: Must be my confusing phrasing. What I meant in the comment was "There is no limit applied if the query parameter is empty or non-positive".
Looking at it now, I really want to rephrase it as well.

build.go
jilleJr: nit: Could you write "Same for the oither `@limit` query params in the other endpoints as well" instead of duplicating the comment? I find it annoying to deal with 5 discussion threads regarding the same thing

I'll mark all but the first thread as resolved

utils_gorm.go
jilleJr: Good suggestion! I don't know either, but I'm assuming when it needs to enlarge the underlying array the lack of pointer makes it "pointless"

utils_gorm.go
jilleJr: :3

artifact.go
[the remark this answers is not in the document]
Alexamakans: How does the `default(100)` part work then?
I assumed it would set it to `100` if the query parameter is not present, i.e. unset.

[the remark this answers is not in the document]
Alexamakans: Ah! Is it just the default value that gets set in the swagger web interface?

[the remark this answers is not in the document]
Alexamakans: And roger that on the nit about duplicating threads 👍🏻

utils_gorm.go
jilleJr: Good question! As the `x BETWEEN ? AND ?` operator in SQL is defined as syntactic sugar for `x >= ? AND x <= ?` then I assumed it didn't matter.

Apparently Postgres has a `x BETWEEN SYMMETRIC ? AND ?`, which automatically flips the operands for you if `min > max`.

Did some manual testing, and both Sqlite3 and PostgreSQL just returned an empty set, which was what I was hoping for.

And if the database dislikes the input anyway, I don't see any problem with wharf-api responding with a "DB read error" problem. Should maybe be a 400 "Bad request" error and not a 502 "Bad gateway" error, but I don't mind.

jilleJr: I.e., I'm OK with leaving it kind of like "undefined behaviour"

artifact.go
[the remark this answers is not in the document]
jilleJr: Yea, the `defaultCommonGetQueryParams` has the Go default of 100, while the `default(100)` is only for Swagger docs.

The binding Gin uses turns empty input string into 0 for an int type. So `?limit=` and `?limit=0` are equivalent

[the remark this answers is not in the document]
jilleJr: I've rephrased it as such:

> Number of results to return. No limiting is applied if empty (`?limit=`) or non-positive (`?limit=0`). Required if `offset` is used.

fredx30 (approved):
I dont like the solution but untill someone comes up with a mroe readable, breifer version of this...

Alexamakans (approved):

PR #111, opened 2021-11-11
Added CreatedAt and UpdatedAt

jilleJr:

## Summary

- Added UpdatedAt and CreatedAt to the following database models:

  - Artifact
  - Branch
  - Build
  - Project
  - Provider
  - TestResultDetail
  - TestResultSummary
  - Token

  Did not add to the follow database models, as I thought it was unnecessary:

  - BuildParam
  - Log
  - Param

## Motivation

GORM supports automatic time tracking fields (<https://gorm.io/docs/models.html#Creating-Updating-Time-Unix-Milli-Nano-Seconds-Tracking>). They are such a "nice to have" feature that only comes in handy if you have it in place, and so difficult to add in later. I don't see any direct uses of it yet, but could be good to make use of later.

Q: Should UpdatedAt and CreatedAt be nullable? Or should we let projects created before this patch have the creation timestamp of "0001-01-01 00:00:00 UTC"?


Alexamakans (approved):

fredx30 (approved):
> Q: Should UpdatedAt and CreatedAt be nullable? Or should we let projects created before this patch have the creation timestamp of "0001-01-01 00:00:00 UTC"?

I would say use default paramters- that seems to be no nullable.

jilleJr:
> > Q: Should UpdatedAt and CreatedAt be nullable? Or should we let projects created before this patch have the creation timestamp of "0001-01-01 00:00:00 UTC"?
> 
> I would say use default paramters- that seems to be no nullable.

So you're suggesting to change from `*time.Time` to `time.Time`, or to leave it be?

jilleJr:
As you approved, I'm assuming "leave it be".

PR #112, opened 2021-11-11
Added RemoteProjectID to Project structs

### LINKED ISSUE #12: Map projects per ID: DB Schema
Projects are currently mapped to their corresponding provider via name. Name can change, such as when a project is moved, but ID does not.

This is such a fundamentally impactful change that we need to be cautious and perhaps make a bunch of migration scripts.

Suggested change would be that the project in Wharfs DB is mapped to the git server and the project ID within that server. For example "gitlab.local" and "193".

---

There has come up a case where the current implementation caused a bug. Especially since the current GitLab importer actually maps the auth token with the project display name. 

See the code (redacted link, it was outdated anyway, but pointed to somewhere in: https://github.com/iver-wharf/wharf-provider-gitlab/blob/master/import.go)
It would make more sense if it used the "path" instead of "name" property of the repo. To show the difference of the two, here a screenshot of GitLabs docs with an example project called "Diaspora Client" but with the path "dispora-client"

image

A project that is damaged by this is Foo Bar, which is accessed via the path "foo-bar" but has the name "Foo Bar" https://gitlab.local/default/foo-bar
When it tries to build docker images it builds them with the destination
`harbor.local/default/Foo Bar:latest`, which is an invalid docker image name.

---

There so much auto-magic and assumed relations tied together here. We need to find a way to migrate the data in as a painless and future-proof way as possible.

Migrating to basing projects off the project ID and server domain instead is a good step in my (@jilleJr's) opinion.
Then we can also import meta data for it such as it's full path so that we later can use that value for default docker image names instead of these home-baked spagetti relations.

---

Steps:
- change db scheme

Alexamakans:

## Summary

Added `RemoteProjectID` to all `Project` models.
Updated `projectconv.go` functions to support `RemoteProjectID`.

## Motivation

Required change to be able to store the remote project ID from remote providers.

---

Closes #12.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -174,6 +174,8 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
   - `test_result_summary.file_name`
   - `token.user_name`
 
+- Added new field `RemoteProjectID` to the project model. (#112)
+
 ## v4.2.0 (2021-09-10)
 
 - Added support for the TZ environment variable (setting timezones ex.
--- a/pkg/model/database/database.go
+++ b/pkg/model/database/database.go
@@ -74,6 +74,7 @@ type Token struct {
 // Preload statements to select the correct field to preload.
 var ProjectFields = struct {
 	ProjectID       string
+	RemoteProjectID string
 	Name            string
 	GroupName       string
 	Description     string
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -174,6 +174,12 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
   - `test_result_summary.file_name`
   - `token.user_name`
 
+- Added new field `RemoteProjectID` to: (#112)
+
+  - `pkg/model/database.Project`
+  - `pkg/model/request.Project`
+  - `pkg/model/response.project`
```
jilleJr, line 181: typo project->Project

Also, I think it's fine to just say "the project model"

```suggestion
- Added new field `RemoteProjectID` to the project model. (#112)
```

pkg/model/database/database.go
```diff
@@ -135,6 +135,7 @@ type Project struct {
 	BuildDefinition string    `gorm:"not null;default:''"`
 	Branches        []Branch  `gorm:"foreignKey:ProjectID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
 	GitURL          string    `gorm:"not null;default:''"`
+	RemoteProjectID string    `gorm:"not null;default:''"`
```
jilleJr, line 138: nit: Suggest placing this just below the ProjectID field. I think it belongs better next to that one

jilleJr (approved):

pkg/model/database/database.go
```diff
@@ -104,26 +105,29 @@ var ProjectFields = struct {
...
+	TokenID:         "token_id",
+	GitURL:          "git_url",
 }
 
 // Project holds data about an imported project. A lot of the data is expected
 // to be populated with data from the remote provider, such as the description
 // and avatar.
 type Project struct {
 	ProjectID       uint      `gorm:"primaryKey"`
+	RemoteProjectID string    `gorm:"not null;default:''"`
```
jilleJr, line 130: Think it'd be wise to add an index to this column.

```suggestion
	RemoteProjectID string    `gorm:"not null;default:'';index:project_idx_remote_project_id"`
```

jilleJr (approved):

pkg/model/database/database.go
jilleJr: Wait, no, we won't be querying on this field. Ignore this

fredx30 (approved):

PR #115, opened 2021-11-12
Added temp fix for CVE-2021-42374 & CVE-20210-42375

jilleJr:
- \[x] I've added a new note in the `CHANGELOG.md` file, according to docs:
  https://iver-wharf.github.io/#/development/changelogs/writing-changelogs

## Summary

- Added `apk add --upgrade` to Dockerfile to update `busybox` package

## Motivation

Reported here:

https://github.com/iver-wharf/wharf-api/security/code-scanning/46?query=ref%3Arefs%2Fheads%2Fmaster

Not yet fixed in upstream alpine Docker image, but a simple `apk add --upgrade` fixes this

PR #116, opened 2021-11-12
Updated .gitignore & .dockerignore

jilleJr:

## Summary

- Changed .gitignore & .dockerignore to be more organized and ignore more types, such as `*.db`, `*.tar`, and `*.tar.gz` files

## Motivation

`*.db` files are created by Sqlite.
`*.tar` files are created here and there, but one use case I came across in #115 is when saving Docker images to disk to be analysed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,13 +1,28 @@
-.git
-.gitignore
-
-.dockerignore
-Dockerfile
-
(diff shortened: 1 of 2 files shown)


Alexamakans (approved):
Nice, looks good.

fredx30 (approved):

PR #117, opened 2021-11-12
Project overrides

### LINKED ISSUE #114: Project overrides
Described by RFC-0024: https://github.com/iver-wharf/rfcs/pull/24

Addition of new table `project_override` which defines almost all of the columns in the `project` table that, if set, will be used in HTTP response body and left untouched by HTTP PUT requests.

jilleJr:

## Summary

- Added project_overrides table

- Added ProjectOverrides model

- Added endpoints for updating, reading, and deleting a project's overrides:

  - `GET /project/{projectId}/override`
  - `PUT /project/{projectId}/override`
  - `DELETE /project/{projectId}/override`

- Changed DB project -> response project mapping function to use the overrides, if available.

- Changed build props generator func to use the overrides, if available.

## Motivation

Based on [RFC-0024](https://iver-wharf.github.io/rfcs/published/0024-project-overrides)

In the RFC, I stated that the table should be named `project_override` (singular), but I'm relying on GORM's automatic naming convention, which instead named the table `project_overrides` (plural).

Closes #114


project.go
```diff
@@ -361,6 +365,136 @@ func (m projectModule) updateProjectHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, resProject)
 }
 
+// getProjectOverridesHandler godoc
+// @id updateProjectOverrides
```
Alexamakans, line 369:
```suggestion
// @id getProjectOverrides
```

Alexamakans (changes requested):
Besides wrong ID, LGTM 👍🏻

pkg/modelconv/modelconv.go
```diff
@@ -3,3 +3,10 @@ package modelconv
 import "github.com/iver-wharf/wharf-core/pkg/logger"
 
 var log = logger.NewScoped("WHARF")
+
+func fallbackString(value, fallback string) string {
+	if value == "" {
```
Pikabanga, line 8: Can it be nil?

pkg/modelconv/modelconv.go
```diff
@@ -3,3 +3,10 @@ package modelconv
 import "github.com/iver-wharf/wharf-core/pkg/logger"
 
 var log = logger.NewScoped("WHARF")
+
+func fallbackString(value, fallback string) string {
```
Pikabanga, line 7: If there's a C# params style parameter in Go it would be cool to be able to send in multiple fallbacks and return the first one with content by looping through them.

utils.go
```diff
@@ -21,6 +21,13 @@ func contains(items []string, value string) bool {
 	return false
 }
 
+func fallbackString(value, fallback string) string {
```
Pikabanga, line 24: duplicated func

Pikabanga (changes requested):

pkg/modelconv/modelconv.go
jilleJr: Nope, strings cannot be nil in Go.

Pointers to strings can be, but `value` here is just a regular string

pkg/modelconv/modelconv.go
jilleJr: Yes there is! As nice use case as that is though, I would like to wait with implementing that until we need it.

utils.go
jilleJr: I had trouble making this function reusable (i.e. exported) as they are used in two separate packages.

I'll make another attempt though

jilleJr: New package: `internal/or`

pkg/modelconv/modelconv.go
jilleJr: I went with variadic params anyway. See the implementation in `internal/or/or.go`

Alexamakans (approved):

internal/or/or.go
```diff
@@ -0,0 +1,13 @@
+// Package or helps with a simple use case: defining fallback values, similar
+// to the null coalescing operator.
+package or
```
fredx30, line 3: Perhaps something like stringUtils? Seeing only `or` in a directory listing is unlikely to come with enough context that it could be considered a descriptive name.

fredx30 (approved):

internal/or/or.go
jilleJr: Want the name to also be short, and this logic could easily be extended for other types so I don't think adding `string` into the package name is needed.

What about `nonzero.String`?

Alexamakans: More alternatives:
`first.String`
`valid.String`: Feels more like it would return a bool, don't particularly like it myself.
`fallback.String`: A bit long but fine I think.

I prefer `first`, `or`, and `fallback`.

jilleJr: More alternatives:
- `firstnonzero.String`
- `nonzero.FirstString`

project.go
```diff
@@ -39,12 +39,23 @@ func (m projectModule) Register(g *gin.RouterGroup) {
 	{
 		project.GET("", m.getProjectListHandler)
 		project.POST("", m.createProjectHandler)
-		project.DELETE("/:projectId", m.deleteProjectHandler)
 		project.PUT("/:projectId", m.updateProjectHandler)
 		project.POST("/:projectId/:stage/run", m.startProjectBuildHandler)
```
Alexamakans: These should be removed I think, seem to be duplicate of those in the `:projectId` group below.
```suggestion
```

Alexamakans (changes requested):

project.go
jilleJr: Correct. Fixed

Alexamakans (approved):

internal/or/or.go
Pikabanga: In SQL it's called `coalesce`

Pikabanga (approved):

internal/or/or.go
jilleJr: @Pikabanga Did not know that! I'm changing to that until someone complaints

Pikabanga (approved):

internal/or/or.go
jilleJr: Noone complained. I'm merging.

PR #118, opened 2021-11-15
Paginated response fixes

jilleJr:

## Summary

- Changed test_result.go to use paginated responses
- Renamed slice field in paginated response to only be "list"

## Motivation

Having the fields be named differently is just a hassle if your client supports any kind of generics. While Go doesn't, we don't want to be an easily avoidable annoyance.

Changing the test_result.go endpoints was brought up in discussion <https://github.com/iver-wharf/wharf-api/pull/109#issuecomment-966894215>, and was done to keep consistency with the other GET endpoints

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
   `"totalCount"` field for the overall query.
 
   By default all these new endpoints use a default limit of 100, but this can
-  be disabled by specifying `?limit=0`. (#109)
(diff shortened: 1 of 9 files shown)


Alexamakans (approved):

fredx30 (approved):
No generics available yet right? #waiting-on-go-2

jilleJr:
> No generics available yet right? #waiting-on-go-2

I actually don't think Go should get generics. #let-go-2-take-their-time #i-hope-it-gets-even-more-delayed

PR #121, opened 2021-11-18
Implemented endpoints from #120

jilleJr:

## Summary

- Added checkProjectExistsByID & checkBuildExistsByID
- Implemented `GET /project/{projectId}/branch`
- Implemented `POST /project/{projectId}/branch`
- Implemented `PUT /build/{buildId}/status`
- Updated "NOT IMPLEMENTED YET" -> "NOT IMPLEMENTED" in deprecated endpoint

## Motivation

PR #120 marked some endpoints as deprecated and added new placeholder endpoints that should be used instead, but only returned "501 Not Implemented".

This PR adds implementation for them.

Closes https://github.com/iver-wharf/wharf-api/milestone/1. Yes, the final effort for RFC-0016

### DIFF
--- a/branch.go
+++ b/branch.go
@@ -67,6 +67,7 @@ func (m branchModule) getProjectBranchListHandler(c *gin.Context) {
 // @tags branch
 // @accept json
 // @produce json
+// @param projectId path uint true "project ID" minimum(0)
 // @param branch body request.Branch true "Branch object"
 // @success 200 {object} response.Branch "Updated branches"
 // @failure 400 {object} problem.Response "Bad request"
--- a/branch.go
+++ b/branch.go
@@ -6,6 +6,7 @@ import (
 	"github.com/iver-wharf/wharf-api/internal/ptrconv"
 	"github.com/iver-wharf/wharf-api/pkg/model/database"
 	"github.com/iver-wharf/wharf-api/pkg/model/request"
+	"github.com/iver-wharf/wharf-api/pkg/model/response"
 	"github.com/iver-wharf/wharf-api/pkg/modelconv"
 	"github.com/iver-wharf/wharf-core/pkg/ginutil"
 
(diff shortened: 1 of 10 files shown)


Alexamakans (approved):

branch.go
```diff
@@ -182,7 +253,7 @@ func createBranchesWithNames(db *gorm.DB, projectID, tokenID uint, branchNames [
 func deleteBranchesByNames(db *gorm.DB, projectID uint, branchNames []string) error {
 	return db.
 		Where(&database.Branch{ProjectID: projectID}, database.BranchFields.ProjectID).
-		Where(database.BranchColumns.Name+" IN ?", branchNames).
+		Where(database.BranchColumns.Name+" IN ?", stringSliceToInterfaces(branchNames)).
 		Delete(&database.Branch{}).
```
jilleJr, line 257: This one was an annoying issue.

In the documentation, it states that you can pass in a slice array: https://gorm.io/docs/query.html#String-Conditions

But with some testing that resulted in this:

```sql
DELETE FROM `branch`
WHERE `branch`.`branch_id` IN (
  "name IN ?",
  ("main","feature/test","feature/test2","feature/fixed-thing","dev")
) AND `branch`.`project_id` = 1
```

Converting the type from `[]string` to `[]interface{}` makes it work.
Will write a note about this in the iver-wharf.github.io docs down in the "Things to keep in mind" section

jilleJr: I've opened a PR on it here:

https://github.com/iver-wharf/iver-wharf.github.io/pull/79

Alexamakans (approved):

branch.go
```diff
@@ -28,27 +29,97 @@ func (m branchModule) Register(g *gin.RouterGroup) {
...
+// @failure 502 {object} problem.Response "Database is unreachable"
 // @router /project/{projectId}/branch [get]
 func (m branchModule) getProjectBranchListHandler(c *gin.Context) {
-	c.Status(http.StatusNotImplemented)
+	projectID, ok := ginutil.ParseParamUint(c, "projectId")
+	if !ok {
+		return
+	}
+	if !checkProjectExistsByID(c, m.Database, projectID, "when fetching list of branches for project") {
+		return
```
fredx30, line 47: I would prefer a recoverable crash/http500 to silently ignored errors. This seems to be missing log/error output.

I find some of the most difficult to debug errors are the ones where something is expected to happen but nothing actually happens followed by a code 0 style exit.

fredx30: There are other similar cases of this further down the file.

jilleJr: I guess the naming is misleading. It does write a `problem.Response` to the HTTP response.

Maybe instead have "ensure" instead of "check"? Like `ensureProjectExistsByID`?

fredx30: Ahh i see what hapens here the funtion that gets called by the function that gets called in the if statement does logging. There is some abstract rule about never doing writes in the check for an if- i think it falls under idempotency. Im not sure logging qualifies here. Atleast in the case of a large application it probably would not. 

The name is probably fine. Although i guess your right- there is something to the word clarifying the boolean type of the return type that describes how much a function does. I guess i would expect a function named `validateProject(...)` (perhaps with a polymorphics signature accepting `int id`, or `Project proj`) to return state valid vs not valid and perhaps do a db read. While i would expect a function such as `isExsistProject(...)` || `isProject(...)`  to be more like a null or type check.

fredx30 (approved):

branch.go
jilleJr: I changed to `validate...` instead. Is that sufficient? Or is it still too unclear?

Alexamakans (approved):

branch.go
fredx30: I like it :)

PR #124, opened 2021-11-22
Changed .wharf-ci.yml for canary builds

jilleJr:

## Summary

- Changed deploy tag in `.wharf-ci.yml` to `:canary`
- Changed to use our internal Docker Hub registry caching proxy in Docker builds
- Fixed Wharf build step name in `.wharf-ci.yml`

## Motivation

Minor tweaks to make the Wharf builds work and be able to build Wharf itself.

~~The image tag name `:unstable` is up for debate. Perhaps `:canary`? Or simply `:latest`?~~ Changed to `:canary`.

Related to https://github.com/iver-wharf/iver-wharf.github.io/issues/80

This is the first PR I'm marking "Ready for review". Let's have all discussions in this PR to begin with, as the other PRs listed in the meta-issue above are basically just the same.

### DIFF
--- a/.wharf-ci.yml
+++ b/.wharf-ci.yml
@@ -7,3 +7,4 @@ build:
         - BUILD_VERSION=${GIT_BRANCH}
         - BUILD_GIT_COMMIT=${GIT_COMMIT}
         - BUILD_REF=${BUILD_REF}
+        - REG=${REG_URL}/hub
--- a/.wharf-ci.yml
+++ b/.wharf-ci.yml
@@ -2,7 +2,7 @@ build:
   wharf-api:
     docker:
       file: Dockerfile
-      tag: unstable
+      tag: canary
       args:
         - BUILD_VERSION=${GIT_BRANCH}
         - BUILD_GIT_COMMIT=${GIT_COMMIT}
--- a/.wharf-ci.yml
+++ b/.wharf-ci.yml
@@ -1,10 +1,9 @@
 build:
-  api:
+  wharf-api:
     docker:
       file: Dockerfile
-      tag: ${GIT_COMMIT},${GIT_TAG},latest
-      append-cert: false
+      tag: unstable
       args:
-        - BUILD_VERSION=${GIT_TAG}
+        - BUILD_VERSION=${GIT_BRANCH}
         - BUILD_GIT_COMMIT=${GIT_COMMIT}
         - BUILD_REF=${BUILD_REF}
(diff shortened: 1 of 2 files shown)


fredx30 (commented):
Whats the use case for this?
Edit: Nvm i think it might be the continuation to the discussion over the earlier named "development" environment. I still think this would be better off tied to a specific branch rather than just tags.

Also the name unstable just seems like code for undesierable. I would preffer canary to unstable if those are the available choices.

Also i think lastest should be only prod build such that the tag moves with the actualy releases.

jilleJr:
> @fredx30 /.../ I still think this would be better off tied to a specific branch rather than just tags.

The planned flow is as follows:

1. Go to Wharf
2. Find project iver-wharf/wharf-api
3. Press start new build
4. Select branch you want deployed
   Screenshot from 2021-11-22 11-24-56

5. Once build has finished, find our "deploy-unstable"/"deploy-canary" project and start a new build there
6. Done

This depends on branches and not on just tags. Or are you referring to the Docker image tag?

.wharf-ci.yml
```diff
@@ -1,10 +1,10 @@
...
-  api:
+  wharf-api:
     docker:
       file: Dockerfile
-      tag: ${GIT_COMMIT},${GIT_TAG},latest
-      append-cert: false
+      tag: canary
       args:
-        - BUILD_VERSION=${GIT_TAG}
+        - BUILD_VERSION=${GIT_BRANCH}
```
Pikabanga, line 7: Should this really be changed?

Pikabanga (changes requested):

.wharf-ci.yml
jilleJr: The idea is not to build released versions, but instead any arbitrary commit from any branch. So the `GIT_TAG` Wharf variable will be empty most of the times, and the `BUILD_VERSION` Docker argument is set to `GIT_BRANCH` mostly just to get some kind of info about where the build is from.

For example, I used these builds and deployed them to our internal staging env, resulting in:

Screenshot from 2021-11-24 09-16-19

In contrast to:

Screenshot from 2021-11-24 09-16-57

Pikabanga (approved):

.wharf-ci.yml
```diff
@@ -1,10 +1,10 @@
 build:
-  api:
+  wharf-api:
```
Alexamakans, line 2: Does this require any changes anywhere else?

Alexamakans (approved):
LGTM 👍🏻

.wharf-ci.yml
jilleJr: No

PR #131, opened 2022-01-20
Preload test result summaries and params when getting list of builds

Alexamakans:

## Summary

Added preloading of test result summaries and params when getting a list of builds.

## Motivation

Required to be able to show the test result list summary (currently generated from test result summaries when getting) in the builds list view in iver-wharf/wharf-web/pull/53 without making an extra request per build.

Params preloaded as well to match getting a single build, and so it can be wrapped in a function.

## Alternative solution

Keep the preloading, but empty the `TestResultSummaries` field after generating the `TestResultListSummary` to save bandwidth when there are many tests. Thoughts?

## Notes

I feel this is more of a bugfix, as I believe the idea was to have `testResultListSummary` be available _especially_ when fetching multiple builds. Hence patch-version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 	https://changelog.md/
 -->
 
+## v5.0.1 (WIP)
+
+- Fixed `testResultListSummary`, `testResultSummary` and `params` not filled
+  when fetching list of builds. (#131)
+
 ## v5.0.0 (2022-01-17)
 
 - BREAKING: Changed module path from `github.com/iver-wharf/wharf-api` to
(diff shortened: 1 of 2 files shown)


jilleJr:
> Keep the preloading, but empty the `TestResultSummaries` field after generating the `TestResultListSummary` to save bandwidth when there are many tests. Thoughts?

Sounds like future improvements if we notice it's an issue. "Premature optimizations=evil" and all that.

But to theorize: The `TestResultSummaries` model only correlates to a single test result file, yes? And the biggest use case we have right now has about 10 files uploaded per build, and the builds are paginated by default at a chunk of 10 builds. So worst case we download 100 test result summaries from the database on each build listing. And then the `TestResultSummaries` model only includes mostly integers (as you're not preloading their associated artifact or build metadata) and a string for the file name, so we're talking <1k bytes or something. Not that heavy on neither the bandwidth, the database, nor the wharf-api.

For now, let's talk optimizations when this becomes an issue. Such as if a user wants to download all builds, then we could have a query parameter for omitting the test result summaries or something. But again, let's consider that later.

jilleJr (approved):

fredx30 (approved):

PR #134, opened 2022-01-28
Added execution engine configs

### LINKED ISSUE #132: Add option for selecting execution engine
Add a query parameter to when starting builds to select the execution engine. Such as:

```http
POST /api/project/{projectId}/build?engine=wharf-cmd
```

The value possible here should come from the configs. Such as:

```yaml
ci:
  defaultEngine: jenkins
  engines:
    jenkins:
      name: Jenkins
      triggerUrl: https://jenkins.local/whatever
      triggerToken: changeit
    wharf-cmd:
      name: wharf-cmd
      triggerUrl: http://wharf-cmd-provisioner
      triggerToken: changeit
```

Needs backwards compatbility, so if defining `ci.triggerUrl` and `ci.triggerToken` (the existing config fields) then that should be equivalent to:

```yaml
ci:
  # old fields:
  triggerUrl: https://jenkins.local/whatever
  triggerToken: changeit

  # are translated to the following:
  defaultEngine: unnamed
  engines:
    unnamed:
      name: Unnamed
      triggerUrl: https://jenkins.local/whatever
      triggerToken: changeit
```

There should also be new endpoints for accessing these endpoints, so the frontend can display a way to select between them:

```http
GET /api/engine
```

Response:

```json
{
  "default": "jenkins",
  "engines": {
    "jenkins": {
      "name": "Jenkins",
      "triggerUrl": "https://jenkins.local/whatever"
    },
    "wharf-cmd": {
      "name": "wharf-cmd",
      "triggerUrl": "http://wharf-cmd-provisioner"
    }
  }
}
```

The response cannot include the tokens. But the URLs are safe to share and can be used to give more context to the user.

jilleJr:

## Summary

- Added config:

  ```yaml
  ci:
    # old way:
    triggerUrl: http://jenkins.local/trigger
    triggerToken: changeit

    # new way:
    engine:
      id: jenkins # defaults to "primary"
      name: Jenkins # defaults to "Primary"
      url: http://jenkins.local/trigger # defaults to ci.triggerUrl
      token: changeit # defaults to ci.triggerToken

    engine2:
      id: wharf-cmd # defaults to "secondary"
      name: wharf-cmd # defaults to "Secondary"
      url: http://wharf-cmd-provisioner/api/trigger
      token: changeit
  ```

- Added response models for engines
- Added endpoint `GET /api/engine`
- Added query parameter `?engine=` to "start build" endpoint
- Added engine ID to build table
- Added engine to build responses via engineLookup
- Added a new problem type: `/prob/api/engine/no-default`

## Motivation

Allows for a more natural transition over to wharf-cmd instead of Jenkins, as the user can specify what execution engine they want to use on a per-build basis.

With such, we can have Jenkins as the primary and default engine until wharf-cmd is feature-complete with our Jenkins implementation. Then we could switch so wharf-cmd is the primary engine and keep Jenkins as a secondary to allow using that one if the wharf-cmd implementation fails.

The config design is a bit weird. The reason is that Viper, the config library we're using via wharf-core, doesn't support arbirary names or indices for environment variables. It can only perform lookups on fields it already recognizes. See #132 for the initial YAML config design using a `map[string]Engine`. This config design limits us for now to only having a primary and a secondary engine, but can be expanded further by adding `ci.engine3`, `ci.engine4`, etc. From the REST API perspective, it's just a list of engine definitions.

## After merge

- Create PR to iver-wharf.github.io for docs on new problem type: `/prob/api/engine/no-default`

---

Closes #132


fredx30 (approved):

Alexamakans (approved):

jilleJr:
I pushed two minor bug fixes. Hope you're OK with me merging after you review again.

PR #138, opened 2022-02-07
Moved linting to Makefile

jilleJr:
## Summary

- Added targets to `Makefile` for linting and other utilities like cleaning
- Removed linting from NPM scripts in `package.json`
- Simplified linting docs

## Motivation

Moves focus from relying on NPM for all linting stuff over to Makefile.

Based on https://github.com/iver-wharf/wharf-cmd/pull/31

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ RUN chmod +x deploy/update-version.sh  \
     && deploy/update-version.sh version.yaml \
     && make swag \
     && CGO_ENABLED=0 go build -o main \
-    && make test
+    && make check
 
 ARG REG=docker.io
 FROM ${REG}/library/alpine:3.14 AS final
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,9 +12,8 @@ ARG BUILD_REF="0"
 ARG BUILD_DATE=""
 RUN chmod +x deploy/update-version.sh  \
     && deploy/update-version.sh version.yaml \
-    && make swag \
-    && CGO_ENABLED=0 go build -o main \
-    && make check
+    && make swag check \
+    && CGO_ENABLED=0 go build -o main
 
 ARG REG=docker.io
 FROM ${REG}/library/alpine:3.14 AS final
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN chmod +x deploy/update-version.sh  \
     && deploy/update-version.sh version.yaml \
     && make swag \
     && CGO_ENABLED=0 go build -o main \
-    && make test
+    && make check
```
Alexamakans, line 17: Would this be possible as well? Removing `make swag` and swapping line 16 and 17.
If not, then why?

```Dockerfile
&& deploy/update-version.sh version.yaml \
&& make check \
&& CGO_ENABLED=0 go build -o main \
```

jilleJr: Yea sure. I think I added it separately to have it explicit that we do both, but merging the two `make` calls will let make to skip the duplicate `swag` target runs.

Dockerfile
```diff
@@ -12,9 +12,8 @@ ARG BUILD_REF="0"
 ARG BUILD_DATE=""
 RUN chmod +x deploy/update-version.sh  \
     && deploy/update-version.sh version.yaml \
-    && make swag \
-    && CGO_ENABLED=0 go build -o main \
-    && make test
+    && make swag check \
+    && CGO_ENABLED=0 go build -o main
```
fredx30, line 16: Whats this change?

Makefile
```diff
@@ -43,6 +69,18 @@ endif
 endif
 	@# This comment silences warning "make: Nothing to be done for 'swag'."
 
-deps:
-	cd .. && go get -u github.com/swaggo/swag/cmd/swag@v1.7.1
-	go mod download
+lint: lint-md lint-go
+lint-fix: lint-fix-md
```
Alexamakans, line 73: Does not contain `lint-fix-go`
```suggestion
lint-fix: lint-fix-md lint-fix-go
```

Alexamakans (changes requested):

Makefile
jilleJr: Got damn it x)

Alexamakans (approved):

Issue #145, opened 2022-02-17
Sqlite foreign keys


jilleJr:
Newer versions of Sqlite supports foreign key constraints, as does the newer versions of the `github.com/mattn/go-sqlite3` package.

Should add this in as we sometimes rely on the constraints to CASCADE DELETE rows, without which we currently get orphan rows in the local `wharf-api.db` database file.

PR #147, opened 2022-02-21
Added gRPC logging stream server

### LINKED ISSUE #130: gRPC server for streaming logs insertions
For faster logs insertions, we add a gRPC API for streaming logs into the wharf-api.

Ex:

```proto
syntax = "proto3";
import "google/protobuf/timestamp.proto";

service Logger {
  rpc CreateLogStream(stream CreateLogStreamRequest) returns (CreateLogStreamResponse);
}

message CreateLogStreamRequest {
  uint64 step_id = 0;
  uint64 log_id = 1;
  string log_message = 2;
  google.protobuf.Timestamp timestamp = 3;
}
```

jilleJr:
- \[x] Wait for #148 to be merged.

## Summary

- Added `api/wharfapi/v1/builds.proto`
- Added Protobuf compiling to `Makefile`
- Added dependency on Go gRPC libraries
- Added multiplexing of gRPC and Gin servers on same port via new dependency `soheilhy/cmux`
- Added gRPC server implementation in `grpc.go`
- Added database log batch injection SQL for sqlite & postgres.
- Moved all HTTP server stuff to `http.go`

## Motivation

The SQL code is what mostly stands out in this PR IMO. What I wanted to do was:

- In a single SQL query do batch insertion of multiple log lines.
- Skip insertion of logs that targets non-existing builds to not cause foreign key violation.
- Get a list back of the logs that were not skipped, so we can publish them via server-sent events (SSE) to any web clients.

In Sqlite this was simple, as you basically do `INSERT OR IGNORE INTO` instead of just `INSERT INTO`, while Postgres does not have such a feature, but I found a solution that instead made use of a `JOIN` statement inside the `INSERT`.

Please look at the test files for the best overview of what the SQL actually looks like.

The Sqlite function isn't really working as we do not have foreign key constraints enabled. However, this is planned in #145.

## Protobuf versioning

I've added the major version to both the path in `api/wharfapi/v5/builds.proto` as well as the Protobuf package name `wharf.api.v5`.

My reasoning here is that it should be the same as the wharf-api version, for simplicity sake. But to be able to keep backwards compatiblity, when jumping to v6 the wharf-api needs to keep supporting v5. In other words, when bumping to v6, we will have the files:

- `api/wharfapi/v5/*.proto`
- `api/wharfapi/v6/*.proto`

For v7, we will have the files:

- `api/wharfapi/v6/*.proto`
- `api/wharfapi/v7/*.proto`

Up for discussion: Should the gRPC protobuf API have a separate versioning? It could, but I think that'd be adding excess confusion.

It does however make it so if we have a breaking change in the HTTP REST API then we also have to bump the gRPC API. Perhaps that's a good thing, but I'm not sure. Input on this is well appreciated.

---

Closes #130


api/wharfapi/v5/builds.proto
```diff
@@ -0,0 +1,48 @@
...
+// database.
+message NewLogLine {
+  // BuildId is the database ID of the build this log line belongs to.
+  uint64 build_id = 1;
+  // LogId is the worker's own ID of the log line. It's unique per build step
+  // for a given build, but may have collisons across multiple steps or builds.
+  uint64 log_id = 2; // not DB value; worker's own ID of the log line
+  // StepId is the worker's own ID of the step. It's unique for a given build,
+  // but may have collisons across multiple builds.
+  uint64 step_id = 3; // not DB value; worker's own ID of the step ID
```
Alexamakans, line 43: Nit: I don't like that we're being forced by this one file to have inconsistent naming in our Go code.
I would prefer we break convention in proto and get `XxxID` for our Go code, like:

```suggestion
  // BuildID is the database ID of the build this log line belongs to.
  uint64 buildID = 1;
  // LogID is the worker's own ID of the log line. It's unique per build step
  // for a given build, but may have collisions across multiple steps or builds.
  uint64 logID = 2; // not DB value; worker's own ID of the log line
  // StepID is the worker's own ID of the step. It's unique for a given build,
  // but may have collisions across multiple builds.
  uint64 stepID = 3; // not DB value; worker's own ID of the step ID
```

main.go
```diff
@@ -63,111 +58,48 @@ func main() {
...
 
-	err = runDatabaseMigrations(db, config.DB.Driver)
+func serve(config Config, db *gorm.DB) {
+	listener, err := net.Listen("tcp", config.HTTP.BindAddress)
 	if err != nil {
-		log.Error().WithError(err).Message("Migration error")
-		os.Exit(3)
+		log.Error().WithError(err).
+			WithString("address", config.HTTP.BindAddress).
+			Message("Failed to bind address.")
```
Alexamakans, line 70: Missing `os.exit` or `return`?

CHANGELOG.md
```diff
@@ -38,6 +38,10 @@ This project tries to follow [SemVer 2.0.0](https://semver.org/).
 - Added query parameter `?engine=ID` to `POST /api/project/{projectId}/build`
   to allow specifying which execution engine to use for the new build. (#134)
 
+- Added gRPC server (on the same port) for injecting logs in a single stream.
+  See [`api/wharfapi/v5/builds.proto`](./api/wharfapi/v5/builds.proto) for full
+  documentation of the API. (#147)
+
```
Alexamakans, line 44: Reminder to add dependencies to the changelog.

Alexamakans (approved):
Good stuff 👍🏻 

Nice to see how simple the multiplexing usage is o:

api/wharfapi/v5/builds.proto
Alexamakans: Actually kinda changed my mind, better to follow convention for proto so it's easier to maintain consistency between protofiles perhaps.

Maybe there are some tags/other ways to change the generated name though?

jilleJr: I'm following the official style guide here https://developers.google.com/protocol-buffers/docs/style#message_and_field_names

I did some digging and found this: https://github.com/alta/protopatch

I've updated this and the wharf-api-client-go PR to make use of it.

I did try out https://github.com/gogo/protobuf but that project is both dying and has lacking support for Protobuf v3. This alta/protopatch solution is only augmenting the official Go protobuf generator instead of replacing it.

main.go
jilleJr: Good call! Fixed.

CHANGELOG.md
jilleJr: Nice catch

api/wharfapi/v5/builds.proto
Alexamakans: Neat 👍🏻

go.mod
```diff
@@ -60,18 +62,22 @@ require (
 	github.com/pelletier/go-toml v1.2.0 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/spf13/afero v1.1.2 // indirect
+	github.com/soheilhy/cmux v0.1.5
```
Alexamakans, line 65: Appears to still be missing from changelog

Alexamakans (approved):

build.go
```diff
@@ -382,6 +383,93 @@ func (m buildModule) createBuildLogHandler(c *gin.Context) {
...
+	return resultLogs, nil
+}
+
+func createLogBatchPostgres(db *gorm.DB, dbLogs []database.Log) ([]database.Log, error) {
+	var result []database.Log
+	q := createLogBatchPostgresQuery(db, dbLogs).Find(&result)
+	return result, q.Error
+}
+
+// TODO: Use database names from pkg/model/database
```
jilleJr, line 412:
```suggestion
```

This TODO was resolved.

PR #150, opened 2022-02-28
Changed build status query to accept arrays

jilleJr:

## Summary

- Changed `status` and `statusId` to be slices instead of single values

## Motivation

This is to allow wharf-cmd-watchdog to query for non-finished builds, as any finished builds are in either status `Completed` or `Failed`, which the API didn't support querying for before this PR.

We use the `multi` collection format, meaning to specify multiple values you specify the query parameter multiple times, such as:

```http
GET /api/build?status=Scheduling&status=Running
```

Because of this, the parameter only gains functionality and doesn't break backwards.

### DIFF
--- a/build.go
+++ b/build.go
@@ -152,8 +152,8 @@ var defaultGetBuildsOrderBy = orderby.Column{Name: database.BuildColumns.BuildID
 // @param gitBranch query string false "Filter by verbatim build Git branch."
 // @param stage query string false "Filter by verbatim build stage."
 // @param isInvalid query bool false "Filter by build's valid/invalid state."
-// @param status query string false "Filter by build status name" enums(Scheduling,Running,Completed,Failed)
-// @param statusId query int false "Filter by build status ID. Cannot be used with `status`." enums(0,1,2,3)
+// @param status query []string false "Filter by build status name" enums(Scheduling,Running,Completed,Failed)
+// @param statusId query []int false "Filter by build status ID. Cannot be used with `status`." enums(0,1,2,3)
 // @param environmentMatch query string false "Filter by matching build environment. Cannot be used with `environment`."
 // @param gitBranchMatch query string false "Filter by matching build Git branch. Cannot be used with `gitBranch`."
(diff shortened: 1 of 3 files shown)


build.go
```diff
@@ -179,8 +179,8 @@ func (m buildModule) getBuildListHandler(c *gin.Context) {
 
 		IsInvalid *bool `form:"isInvalid"`
 
-		Status   *string `form:"status"`
-		StatusID *int    `form:"statusId" binding:"excluded_with=Status"`
+		Status   []string `form:"status"`
+		StatusID []int    `form:"statusId" binding:"excluded_with=Status"`
```
Alexamakans, line 183: nit: Maybe exclude `status` if both are present as `statusID` translates more directly to the database values, and requires less computation? Similar to e.g. `environmentMatch` vs `environment`.

Alexamakans (approved):
Nice 👍🏻

build.go
jilleJr: I would actually argue the other way around. Much better IMO to only expose the string enum variants as they are much easier to work with in the long run and debug.

The computational effort you're talking about is just a `strings.ToLower` and a switch statement. Premature optimization.

I've created an issue on the subject: https://github.com/iver-wharf/wharf-api/issues/152

Alexamakans: Main argument was that it should drop the one that is less directly equivalent to the database value, if we're changing that then no objections from me 👍🏻 Agree that storing as enum is nicer as well 👍🏻

jilleJr: Was actually not suggesting changing the value stored in the database, though that's an option as well. Just that exposing an integer enum through a REST API is more tedious to deal with than a string enum. Same argument applies to the database layer, so maybe we should change it there to. In Go we can quite easily change it to be a string and keep backward compatibility by implementing the [`driver.Valuer`](https://pkg.go.dev/database/sql/driver#Valuer) interface

Alexamakans: Ah I see.
Yeah I think having it stored as a string enum in the database as well would be good for the same reasons.

Issue #154, opened 2022-02-28
Add cancelling build support


jilleJr:
- Add status BuildCancelled
- Add endpoint to cancel build
  - if engine is wharf-cmd: tell wharf-cmd-provisioner to cancel the build
  - otherwise: say cancelling builds isnt supported

Maybe need to extend the engines type to also allow declaring the engine type, with an enum of `jenkins-generic-webhook-trigger` and `wharf-cmd`

PR #156, opened 2022-02-28
Added worker ID field to build

jilleJr:

## Summary

- Added worker ID to build
- Added config.CI.Engine.API enum to declare what kind of engine API it is.
- Parse worker ID from build trigger response if engine is wharf-cmd.v1

## Motivation

Adds worker ID support.

wharf-cmd-watchdog needs this value, and we need this value later in #154 when cancelling builds.

Based on [RFC-0025: wharf-cmd provisioning](https://iver-wharf.github.io/rfcs/published/0025-wharf-cmd-provisioning#wharf-api-changes)


jilleJr:
Wow the CodeQL scan noticed through 20 different nested function calls that we used `c.GetQuery("stage")` all the way down to when we construct the trigger URL where we do `fmt.Sprintf("%s&%s=%s", ...)`

https://github.com/iver-wharf/wharf-api/pull/156/checks?check_run_id=5361232036

However it failed to notice that we did pass the value through `url.QueryEscape`, so I marked it as "false positive"

Alexamakans (approved):
Good stuff 👍🏻

PR #157, opened 2022-02-28
Fixed URL query building

jilleJr:
## Summary

- Changed the trigger URL building to use proper URL building.

## Motivation

Less prone to glitches and security vulnerabilities.

For example, if the engine URL before had a fragment, then it ruined the URL building:

```
http://jenkins.local/trigger#foo
// got turned into:
http://jenkins.local/trigger#foo?token=abc&stage=def&environment=etc
```

While now it correctly edits the correct part:
```
http://jenkins.local/trigger#foo
// now gets turned into:
http://jenkins.local/trigger?token=abc&stage=def&environment=etc#foo
```

### DIFF
--- a/build.go
+++ b/build.go
@@ -773,13 +773,13 @@ func triggerBuild(dbJobParams []database.Param, engine CIEngineConfig) (string, 
 	q.Set("token", engine.Token)
 	u.RawQuery = q.Encode()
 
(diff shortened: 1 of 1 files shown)


Alexamakans (approved):
Nice 👍🏻

See Codacy notice
> var redactedUrl should be redactedURL

PR #166, opened 2022-03-07
Fixed token always getting redacted

jilleJr:

## Summary

- Fixed actual trigger URL used also getting censored

## Motivation

Apparently you cannot make a copy by doing `&(*u)`.

Doesn't matter much as we can just use the dereferenced variant.

Also changed the redacted text as `*****` gets logged as `%2A%2A%2A%2A%2A`, while `~redacted~` isn't escaped.

### DIFF
--- a/build.go
+++ b/build.go
@@ -912,7 +912,7 @@ func triggerBuild(dbJobParams []database.Param, engine CIEngineConfig) (string, 
 	u.RawQuery = q.Encode()
 
 	redactedURL := *u
-	q.Set("token", "*****")
+	q.Set("token", "~~redacted~~")
 	redactedURL.RawQuery = q.Encode()
 
 	log.Info().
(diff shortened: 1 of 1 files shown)


build.go
```diff
@@ -912,15 +911,17 @@ func triggerBuild(dbJobParams []database.Param, engine CIEngineConfig) (string,
...
+	redactedURL.User = nil
+	q.Set("token", "~~redacted~~")
 	redactedURL.RawQuery = q.Encode()
 
 	log.Info().
 		WithString("method", "POST").
 		WithString("url", redactedURL.Redacted()).
 		Message("Triggering build.")
 
+	log.Info().WithString("url", u.String()).Message("Real URL.")
```
Alexamakans, line 924:
```suggestion
```

Alexamakans (changes requested):

Alexamakans (approved):

PR #172, opened 2022-03-30
Added go-typ

jilleJr:

## Summary

- Added go-typ v3.0.1
- Removed unused contains func
- Replaced `internal/set` with `typ/pkg/sets.Set`
- Changed func to generic
- Replaced `internal/coalese` with `typ.Coal()`
- Changed to use `chans.RecvQueued()`

## Motivation

Simplifies some code by using existing libraries. This way if we use the same libs throughout wharf then it's easier to understand, in comparison to where all repos has their own custom code for stuff


Alexamakans (approved):
👍🏻

PR #173, opened 2022-03-30
Replaced strings.Title with cases.Title

jilleJr:
## Summary

- Changed from `strings.Title` to `cases.Title`

## Motivation

[`strings.Title`](https://pkg.go.dev/strings#Title) has the following deprecation warning in Go 1.18:

```go
// Deprecated: The rule Title uses for word boundaries does not handle Unicode
// punctuation properly. Use golang.org/x/text/cases instead.
```


Alexamakans:
Ignoring `Added dependency ...` changelog entry for this because it just went from indirect -> direct?

jilleJr:
> Ignoring `Added dependency ...` changelog entry for this because it just went from indirect -> direct?

Good note, it should be added to CHANGELOG.md then.

Alexamakans (approved):
👍🏻

PR #177, opened 2022-05-03
Updated Makefile

jilleJr:
## Summary

- Updated Makefile

## Motivation

Same structure as the other repo's

Originally based on https://github.com/iver-wharf/wharf-cmd/pull/128

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,25 +1,43 @@
-.PHONY: install tidy deps check \
-	docker docker-run serve swag-force swag proto \
-	lint lint-md lint-go lint-proto \
-	lint-fix lint-fix-md lint-fix-go lint-fix-proto
 
 commit = $(shell git rev-parse HEAD)
 version = latest
 
 ifeq ($(OS),Windows_NT)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -87,14 +114,10 @@ lint-fix-md:
...
 	revive -formatter stylish -config revive.toml ./...
 
 lint-fix-go:
-	goimports -d -w $(shell git ls-files "*.go")
-
-lint-proto:
-	protolint lint api/wharfapi
-
-lint-fix-proto:
-	protolint lint -fix api/wharfapi
```
Alexamakans, line 100: Were there issues with this?
If not then maybe better to keep it in and add it to iver-wharf/wharf-cmd/pull/128 as well instead?

jilleJr: Oops, I mistakenly removed it in my copy-pasting frenzy.

Adding protolint to wharf-cmd should be done via dedicated issue: https://github.com/iver-wharf/wharf-cmd/issues/16

Alexamakans (approved):

PR #180, opened 2022-05-04
Simplified log insertion

jilleJr:

## Summary

- Removed the batch insertion logic into the database.
- Removed the `time.Sleep`... Silly me forgetting to remove that before

## Motivation

This reduces the complexity of wharf-api a lot, while giving up some performance.

It was a premature optimization anyway, and in the end it resulted in glitchy behaviour. We can look back at this PR later if we want a reference implementation later to add it back, but for now we stick with the one-SQL-query-per-insertion logic, which is much faster than we need anyway.


grpc.go
```diff
@@ -29,98 +26,53 @@ func serveGRPC(listener net.Listener, db *gorm.DB) {
...
-		lines := recvQueuedAtLeastOne(logReqChan, bufferSize)
-		if len(lines) == 0 {
+		line, err := stream.Recv()
+		if err == io.EOF {
 			break
+		} else if err != nil {
+			return err
+		} else if line == nil {
+			log.Warn().Message("Received nil message, skipping.")
+			//return status.Error(codes.InvalidArgument, "received nil message")
```
Alexamakans, line 38: Commented out code
```suggestion
```

grpc.go
```diff
@@ -29,98 +26,53 @@ func serveGRPC(listener net.Listener, db *gorm.DB) {
...
-				LogID:     dbLog.LogID,
-				BuildID:   dbLog.BuildID,
-				Message:   dbLog.Message,
-				Timestamp: dbLog.Timestamp,
-			})
-		}
-		logsInserted += uint64(len(createdLogs))
-		time.Sleep(time.Second)
+		log.Debug().WithUint("logId", createdLog.LogID).
+			Message("Inserted logs into database.")
```
Alexamakans, line 66: Only one is inserted now.
```suggestion
			Message("Inserted log into database.")
```

Alexamakans (approved):

